PR #16, opened 2020-09-01
make start and endts optional in slam

fredrikjacobson:
Slam producer no longer requires startTs and endTs.
Change them into optional


Ragge-dev (approved):
I Approve of these changes! :shipit:

PR #20, opened 2020-11-13
Be 987 1142

fredrikjacobson:
Switch to using internal ids `UUIDs` for invalidate input
Add `get_inputs` endpoint


rmeertens:
Nice! Thanks a lot!

PR #26, opened 2020-11-30
Input api documentation

fredrikjacobson:
First stab at readme - to be deployed.

Minor convenience methods to lower the bar for creating inputs.

### DIFF
--- a/annotell-input-api/README.md
+++ b/annotell-input-api/README.md
@@ -41,7 +41,7 @@ images_files = IAM.ImagesFiles(images)
 folder = Path("/home/user_name/example_path")
 ```
 
-**We also need to create a** _**SceneMetaData**_ **object.** The only required field the external id string which the client can use to identify the input. We can also add a _**SourceSpecification**_ object which defines which source that should be shown first, the _source_order_, or a mapping of source names to a prettier name version displayed in the UI.
+**We also need to create a** _**SceneMetaData**_ **object.** The only required field is the external id string which the client can use to identify the input. We can also add a _**SourceSpecification**_ object which defines which source that should be shown first, the _source_order_, or a mapping of source names to a prettier name version displayed in the UI.
 
 ```python
 source_spec = IAM.SourceSpecification(source_to_pretty_name={"CAM1": "FC", "CAM2": "BC"},
--- a/annotell-input-api/README.md
+++ b/annotell-input-api/README.md
@@ -28,9 +28,9 @@ client.list_projects()
 
 ## Create Images
 
-Images inputs can be created from python via the _upload_and_create_images_input_job_ method. An important feature here is the ability to add a _**dryrun**_ parameter to the method call, this will make a call to the inputApi but only run the validation steps and not create any inputs.
+Images inputs can be created from python via the _upload_and_create_images_input_job_ method. An important feature here is the ability to add a _**dryrun**_ parameter to the method call, this will make a call to the Input API but only run the validation steps and not create any inputs.
 
-**We start out by creating a representation of our images.** The representation consists of the image name \(excluding the path to the image\) and the source of the image. In this case, we want to create a scene consisting of two image _image1_. We also specify a folder where the image is located.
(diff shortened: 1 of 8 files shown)


annotell-input-api/annotell/input_api/input_api_model.py
```diff
@@ -85,6 +85,13 @@ def to_dict(self):
         return dict(images=[image.to_dict() for image in self.images],
                     pointClouds=[pc.to_dict() for pc in self.point_clouds])
 
+class PointCloudFiles(RequestCall):
+    def __init__(self, point_clouds: List[Image]):
```
Ragge-dev, line 89: Probably shouldn't be "List[Image]" here

annotell-input-api/examples/point_cloud.py
```diff
@@ -0,0 +1,25 @@
+from annotell.input_api.input_api_client import InputApiClient
+from annotell.input_api import input_api_model as IAM
+from pathlib import Path
+
+print("Creating Point Cloud Input...")
+
+
+# Creating Point Cloud Object
+# Support for .csv, .pcd or .npy
```
Ragge-dev, line 9: Are we supporting .npy files? If so shouldn't we convert them to potree?

annotell-input-api/annotell/input_api/input_api_client.py
```diff
@@ -293,6 +302,79 @@ def create_inputs_point_cloud_with_images(self, folder: Path,
...
+        """Create point clouds"""
+
+        js = dict(
+            files=point_clouds.to_dict(),
+            internalId=internal_id,
+            metadata=metadata.to_dict())
+
+        return self._post_input_request('pointclouds', js, project=project, batch=batch, input_list_id=input_list_id, dryrun=dryrun)
+
+    def create_inputs_point_clouds(self, folder: Path,
```
Ragge-dev, line 322: Sooner rather than later I think that we should align so all the functions that create a input have the same "kind" of name, right now it is a bit scattered. I'll create a jira

annotell-input-api/examples/point_cloud.py
fredrikjacobson: according to docs we do, but from `inputValidationService` case "pcd" | "csv" | "las" | "laz" => true we don't
ill edit

annotell-input-api/annotell/input_api/input_api_client.py
fredrikjacobson: Yes, I believe the right time is when we remove `folder` from the interface and just use resourceIds. At that time, we can rename all methods to "create_xxx_input" and make the uploading a minor detail if a local file is supplied.

annotell-input-api/examples/point_cloud.py
fredrikjacobson: god I can't find where I read we had support for .npy. removing it..

annotell-input-api/README.md
```diff
@@ -1,127 +1,420 @@
...
+```python
+image1 = "filename1.jpg"
+image2 = "filename2.jpg"
+images = [IAM.Image(filename=image1, source="CAM1"),
+          IAM.Image(filename=image2, source="CAM2")]
+images_files = IAM.ImagesFiles(images)
+folder = Path("/home/user_name/example_path")
+```
+
+**We also need to create a** _**SceneMetaData**_ **object.** The only required field the external id string which the client can use to identify the input. We can also add a _**SourceSpecification**_ object which defines which source that should be shown first, the _source_order_, or a mapping of source names to a prettier name version displayed in the UI.
```
Ragge-dev, line 44: Ändrade bara lite i första raden, "The only required field _is_ the external id
```suggestion
**We also need to create a** _**SceneMetaData**_ **object.** The only required field is the external id string which the client can use to identify the input. We can also add a _**SourceSpecification**_ object which defines which source that should be shown first, the _source_order_, or a mapping of source names to a prettier name version displayed in the UI.
```

Ragge-dev (approved):
Nice, supersnyggt! Missuppfattade o tänkte att github.io koden fanns här med, ja kollar på den någon gång när ja känner för det

rmeertens:
Amazing job, this really helps :D

PR #36, opened 2021-02-02
re-directed get-calibration path

Ragge-dev:
Re-directed get calbration path to point towards v0 route


fredrikjacobson (approved):
Do you want to push a new version?

Suppose that could be a good idea to keep things up to date? in that case bump version and add a change log.

PR #38, opened 2021-02-16
Added support for cameras inputs

Ragge-dev:
Edit:Marko completely hijacked my PR!! :sob: The new, final PR, for adding cameras as input type can now be found here:
https://github.com/annotell/annotell-python/pull/40


annotell-input-api/annotell/input_api/resources/calibration/calibration.py
```diff
@@ -2,16 +2,19 @@
...
 
 class CalibrationResource(InputAPIResource):
 
-    def get_calibration(self, id: Optional[int] = None, external_id: Optional[str] = None
-                             ) -> Union[List[IAM.CalibrationNoContent], List[IAM.CalibrationWithContent]]:
+    def get_calibration(
+        self,
+        id: Optional[int] = None,
+        external_id: Optional[str] = None
+    ) -> Union[List[CalibrationModel.CalibrationNoContent], List[CalibrationModel.CalibrationWithContent]]:
```
fredrikjacobson, line 17: I think method overloading might be a cleaner interface here where the listing is done without arguments and the other case is a separate method with optional args.

Also, calibration ID is now a UUID when working with v1 inputs if I recall correctly

annotell-input-api/annotell/input_api/resources/input/cameras.py
```diff
@@ -0,0 +1,49 @@
...
+               dryrun: bool = False) -> Optional[InputModel.CreateInputJob]:
+        """
+        Upload files and create an input of type ``cameras``.
+
+        :param cameras: class containing 2D resources that constitute the input
+        :param project: project to add input to
+        :param batch: batch, defaults to latest open batch
+        :param input_list_id: input list to add input to (alternative to project-batch)
+        :param dryrun: If True the files/metadata will be validated but no input job will be created.
+        :returns CreateInputJobResponse: Class containing id of the created input job, or None if dryrun.
```
fredrikjacobson, line 29: new classname, `InputJob`?

annotell-input-api/annotell/input_api/resources/input/cameras.py
```diff
@@ -0,0 +1,49 @@
...
+
+        The files are uploaded to annotell GCS and an input will be created shortly after submission.
+        """
+
+        self._set_sensor_settings(cameras)
+
+        # We need to set job-id from the response
```
fredrikjacobson, line 36: this line is no longer true since we switched to two phase commit

annotell-input-api/examples/cameras.py
```diff
@@ -0,0 +1,45 @@
...
+
+# Create calibration
+calibration_spec = create_calibration_spec(
+    "Collection 2020-06-16", [], ["RFC01", "RFC02", "RFC03"])
+created_calibration = client.calibration.create_calibration(calibration_spec)
+
+camera_settings = InputModel.CameraSettings(width=1920, height=1080)
+sensor_specification = InputModel.SensorSpecification(sensor_settings=dict(RFC01=camera_settings,
+                                                                           RFC02=camera_settings,
+                                                                           RFC03=camera_settings))
```
fredrikjacobson, line 25: Laurell had a good suggestion to switch from `dict()` to 

```
{
"RFC01": camera_settings
}
```

(Especially in the docs). Or preferably have the sensornames as a variable defined in one place to avoid validation errors that calibration does not match frames.

fredrikjacobson (commented):
Quite a lot going on. a few comments and thoughts.
Overall I like where we are heading with this, will have a look into ergonomics from a user standpoint as well. 

Lets discuss some naming and API:s

annotell-input-api/annotell/input_api/model/input/resource/resource.py
```diff
@@ -0,0 +1,13 @@
+from pathlib import Path
+from abc import ABC
+from dataclasses import dataclass
+
+
+@dataclass
+class Resource(ABC):
+    filename: str
+    resource_id: int
```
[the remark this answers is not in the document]
cotramarko: Yes, you are right

annotell-input-api/annotell/input_api/model/input/resource/video.py
```diff
@@ -1,21 +1,22 @@
...
 
 
 @dataclass
-class VideoFrame:
-    video_timestamp: int
+class VideoFrame(Resource):
     filename: str
     resource_id: Optional[str] = None
     sensor_name: str = camera_sensor_default
+    video_timestamp: int = -1  # TODO: This is due to how dataclasses work, must be named :/
```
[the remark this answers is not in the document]
cotramarko: Yepp, didn't have time to dig into it more

annotell-input-api/annotell/input_api/model/input/input_job.py
```diff
@@ -0,0 +1,13 @@
+from dataclasses import dataclass
+from typing import Dict
+from annotell.input_api.model.abstract.abstract_models import Response
+
+
+@dataclass
+class CreateInputJob(Response):
```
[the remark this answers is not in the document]
cotramarko: Agree

annotell-input-api/annotell/input_api/model/input/input_batch.py
```diff
@@ -1,11 +1,20 @@
 from dataclasses import dataclass
 from datetime import datetime
-from annotell.input_api.model.enums import InputBatchStatus
+from enum import Enum
 from annotell.input_api.util import ts_to_dt
+from annotell.input_api.model.abstract.abstract_models import Response
+
+
+class InputBatchStatus(str, Enum):
+    PENDING = 'pending'
```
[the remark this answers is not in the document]
cotramarko: True, agree

annotell-input-api/annotell/input_api/model/annotation/export_annotation.py
```diff
@@ -0,0 +1,15 @@
...
+class ExportAnnotation(Response):
+    annotation_id: int
+    export_content: dict
+
+    @staticmethod
+    def from_json(js: dict):
+        return ExportAnnotation(
+            annotation_id=int(js["annotationId"]),
+            export_content=js["exportContent"]
+        )
```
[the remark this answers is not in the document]
cotramarko: Yeah it's a mess, was considering of renaming it to just `Annotation(id, content)` but left it as is since as you say, we have a lot to do here

annotell-input-api/annotell/input_api/model/calibration/calib.py
```diff
@@ -0,0 +1,63 @@
...
+@dataclass
+class CalibrationSpec:
+    external_id: str
+    calibration: Calibration
+
+    def to_dict(self):
+        return {
+            'externalId': self.external_id,
+            'calibration': self.calibration.to_dict()
+        }
```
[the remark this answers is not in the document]
cotramarko: Agree

annotell-input-api/annotell/input_api/model/calibration/calib.py
```diff
@@ -0,0 +1,63 @@
...
+    calibration: Mapping[str, dict]
+
+    @staticmethod
+    def from_json(js: dict):
+        return CalibrationWithContent(
+            id=js["id"],
+            external_id=js["externalId"],
+            created=ts_to_dt(js["created"]),
+            calibration=js["calibration"]
+        )
```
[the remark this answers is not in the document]
cotramarko: Agree

annotell-input-api/annotell/input_api/model/calibration/parts.py
```diff
@@ -79,7 +85,7 @@ def get_camera_type(self) -> CameraType:
```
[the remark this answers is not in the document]
cotramarko: Haha why didn't I think of that???? Seriously, I spent like 10 mins trying to think of a good name and then just gave up and called it "parts" as a placeholder name :P

annotell-input-api/annotell/input_api/model/input/cameras/cameras.py
```diff
@@ -0,0 +1,27 @@
...
+                    sensorSpecification=self.sensor_specification.to_dict(),
+                    externalId=self.external_id)
+
+    def get_local_resources(self) -> List[Image]:
+        resources = []
+        for resource in self.frame.images:
+            if resource.resource_id is None:
+                resources.append(resource)
+
+        return resources
```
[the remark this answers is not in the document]
cotramarko: Agree

annotell-input-api/annotell/input_api/resources/calibration/calibration.py
cotramarko: Good point

annotell-input-api/annotell/input_api/resources/input/cameras.py
cotramarko: Yes, I completely forgot about docsstrings

cotramarko:
A lot of great comments @fredrikjacobson, agree on all the stuff you've raised. I do realise that I went loco on this and basically redid a lot of the structure, and I have no doubt missed to update stuff and some of the names are wacko. We'll have to rethink and rename a couple of things.

Looking forward to see what @Ragge-dev thinks about this :)

annotell-input-api/annotell/input_api/model/input/resource/video.py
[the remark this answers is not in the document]
Ragge-dev: Feel like I'm missing something why are we using a default-value for `video_timestamp`? Also I feel reluctant to have comments like this one in the client

annotell-input-api/annotell/input_api/model/input/cameras_sequence/cameras_sequence.py
```diff
@@ -0,0 +1,29 @@
...
+from annotell.input_api.model.input.resource.image import Image
+from annotell.input_api.model.input.resource.video import VideoFrame
+from annotell.input_api.model.input.abstract.camera_sensor import CameraSensor
+
+
+@dataclass
+class CamerasSequence(CameraSensor):
+    external_id: str
+    frames: List[Frame]
+    sensor_specification: SensorSpecification
```
Ragge-dev, line 15: How I think about inheritance is that if X inherits Y, then X **is** a Y and also a subtype of Y. To me it is a bit strange that `CameraSequence` should be a subType of `CameraSensor`, since a `CameraSequence` is not a `CameraSensor`. To me `CameraResource` or `CameraInput` makes more sense, if it is a CameraInput it needs a sensor_specification.

annotell-input-api/annotell/input_api/model/input/lidars_and_cameras_sequence/frame.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+@dataclass
+class Frame:
+    frame_id: str
+    relative_timestamp: int
+    point_clouds: List[PointCloud] = field(default_factory=list)
+    images: List[Image] = field(default_factory=list)
+    video_frames: List[VideoFrame] = field(default_factory=list)
+
```
Ragge-dev, line 15: I am very scared of having four different classes all named `Frame`. Even in Scala I think it is easy to mix up Models with the same name in different directories (e.g. when using auto-complete functionality) and in an untyped language like python, since you won't necessarily realize that you are using the wrong type even at run time, I am convinced this will cause trouble.

I would like the name to be more explicit so that you at a glance could deduce what sort of Frame it is. Maybe using acronyms, like **LCSFrame**, **CFrame**, **CSFrame**? A name like **LidarAndCamerasSequenceFrame** might be long.

cotramarko: Hehe, initially they were actually names what you wrote, `FrameLCS` and so on :P I was a bit divided around that, since it felt really neat to have the same name and instead rely on a "namespaces"-concept via imports instead. But yeah, this can mess things up.... I mean, in some sense, since we're dealing with Python and the best you get is type-hints you could still end up with clients supplying a FrameLCS object to a method that requires a FrameCS objekt, and the code could still end up running

annotell-input-api/examples/cameras.py
```diff
@@ -0,0 +1,45 @@
...
+    external_id="input1",
+    frame=CamerasModel.Frame(
+        images=[
+            ResourceModel.Image("~/Downloads/img_RFC01.jpg", sensor_name="RFC01"),
+            ResourceModel.Image("~/Downloads/img_RFC01.jpg", sensor_name="RFC02"),
+            ResourceModel.Image("~/Downloads/img_RFC01.jpg", sensor_name="RFC03")
+        ]
+    ),
+    sensor_specification=sensor_specification
+)
```
Ragge-dev, line 37: I have come around and think it is quite nice that you added this extra level of hierarchy for all inputs to use (the frame concept). It gives us a recipe to use for all inputs, as long as we push the concept in the docs as well! :)

annotell-input-api/annotell/input_api/model/input/lidars_and_cameras_sequence/frame.py
```diff
@@ -0,0 +1,23 @@
...
+from annotell.input_api.model.input.resource.point_cloud import PointCloud
+
+
+@dataclass
+class Frame:
+    frame_id: str
+    relative_timestamp: int
+    point_clouds: List[PointCloud] = field(default_factory=list)
+    images: List[Image] = field(default_factory=list)
+    video_frames: List[VideoFrame] = field(default_factory=list)
```
Ragge-dev, line 14: How about just naming this video? As I see it a video is basically a list of video frames.

Ragge-dev (commented):
I like the direction you have steered the client towards @cotramarko, I would be happy to use a lot of this in v1. As Fredrik mentioned I think it would be nice to discuss some names though.

More importantly I think we need to discuss our Way of Working, since I think this probably should have been 3 different PR:s (2 for adding new input types, 1 for refactoring). The part I like the least is that I am know dependent on all of your changes, and can't merge before a large scale refactoring has been discussed/approved. 

I will create a new PR with only my changes, without the refactoring, and merge that separately. I hand over the ownership of this PR to you!

annotell-input-api/annotell/input_api/model/input/cameras/cameras.py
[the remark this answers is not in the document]
Ragge-dev: Talked with Fredrik and we just realized that get_local_resources is not used anymore for v1, the Scala part will return what files that need to be uploaded (that is, whatever is missing a resourceId)

annotell-input-api/annotell/input_api/model/input/lidars_and_cameras_sequence/frame.py
Ragge-dev: Yes, I see your point of view as well, and I think it would be good to discuss it some more before we decide. Your way is more "neat" but using a name like FrameCS/FrameC makes it easier imo to use the correct object.

annotell-input-api/annotell/input_api/model/input/cameras_sequence/cameras_sequence.py
cotramarko: Yes, what you're saying makes sense :) Will change it to CameraInput

PR #45, opened 2021-03-03
fixed bug in Input API client

Ragge-dev:
Solves https://annotell.atlassian.net/browse/INT-94

the code setting the camera_settings (in case it is None) did not work due to all our restructuring. Easiest way to get dataclass to work with default values was to create both a CameraFrame and a SequenceFrame class


fredrikjacobson (commented):
Great. 

I'm also considering dropping this in favor of explicitly adding width and height. Especially since we don't have it working for videos.

Ragge-dev:
> Great.
> 
> I'm also considering dropping this in favor of explicitly adding width and height. Especially since we don't have it working for videos.

I would not object to this.

PR #53, opened 2021-03-26
Update project.py

alydh-annotell:


annotell-input-api/annotell/input_api/resources/project/project.py
```diff
@@ -19,7 +19,7 @@ def get_projects(self) -> List[ProjectModel.Project]:
         json_resp = self.client.get("v1/projects")
         return [ProjectModel.Project.from_json(js) for js in json_resp]
 
-    def get_project_batches(self, project: str) -> List[ProjectModel.Project]:
+    def get_project_batches(self, project: str) -> List[ProjectModel.ProjectBatch]:
```
Ragge-dev, line 22: ProjectModel seems like the natural place for ProjectBatch to be, but it is actually in the input model in input_batch.py. I do think though that we should move it to `annotell.input_api.model.project`

Ragge-dev:
I Added a test and moved ProjectBatch to the project model

annotell-input-api/annotell/input_api/resources/project/project.py
fredrikjacobson: agree @Ragge-dev, 

will you merge?

PR #56, opened 2021-04-06
New example with cameras_seq

Ragge-dev:
Solves https://annotell.atlassian.net/browse/INT-166


cotramarko (changes requested):
In terms of code it looks good :) 

However, I think the examples are a bit confusing right now wrt the example files used. Both examples state that 
```
    sensor1 = "RFC01"
    sensor2 = "RFC02"
```

But in the `frames` list we use the same images/video for both sensors, e.g. 

```
video_frames=[
                    InputModel.VideoFrame(
                        filename="./examples/resources/video_RFC01.mp4",
                        sensor_name=sensor1,
                        video_timestamp=0),
                    InputModel.VideoFrame(
                        filename="./examples/resources/video_RFC01.mp4",
                        sensor_name=sensor2,
                        video_timestamp=0),
                ]
```

I suggest adding more example files, so that it's a bit more clear that different sensors will require different files

Ragge-dev:
> In terms of code it looks good :)
> 
> However, I think the examples are a bit confusing right now wrt the example files used. Both examples state that
> 
> ```
>     sensor1 = "RFC01"
>     sensor2 = "RFC02"
> ```
> 
> But in the `frames` list we use the same images/video for both sensors, e.g.
> 
> ```
> video_frames=[
>                     InputModel.VideoFrame(
>                         filename="./examples/resources/video_RFC01.mp4",
>                         sensor_name=sensor1,
>                         video_timestamp=0),
>                     InputModel.VideoFrame(
>                         filename="./examples/resources/video_RFC01.mp4",
>                         sensor_name=sensor2,
>                         video_timestamp=0),
>                 ]
> ```
> 
> I suggest adding more example files, so that it's a bit more clear that different sensors will require different files

Good point, I'll add this for all other examples as well while I'm at it.

cotramarko (approved):
Fantastic, great work! :)

PR #57, opened 2021-04-07
Int 165

cotramarko:
Restructured the documentation.

1. Added a new section called `Overview` which discusses the input format, all the parameters, sequences as well as video vs images. 
2. Renamed the `General` section to `Working with Inputs`, felt like a better name
3. Added an expanded section under `Working with Inputs`->`Downloading Annotations` with example code on how to combine `get_inputs` with `download_annotations`
4. Removed construction-signs and associated input types like `Lidars` and `LidarsSeq` since it's not clear if we'll even have these

### DIFF
--- a/docs-src/docs/input-api/inputs/cameras_seq.md
+++ b/docs-src/docs/input-api/inputs/cameras_seq.md
@@ -2,39 +2,8 @@
 title: CamerasSeq
 ---
 
-A `CamerasSeq` input consists of a sequence of camera images. Unlike single-frame input types you also have to specify the sequential relationship between the _frames_, where each frame consists on 1-8 camera images.
+A `CamerasSeq` input consists of a sequence of camera images, where each frame can contain between 1-8 images from different sensors. For more documentation on what each field corresponds to in the `CamerasSeq` object please check the section related to [Input Overview](/docs/input-api/overview).
 
-## Creating a list of frames
-
-The sequential relationship is expressed via a list of `Frame` objects. This representation expresses the ordering of the frames, but it does not include the _relative temporal_ relationship between the frames in the list.
-
-```python
-frame_1 = IAM.Frame(...)
-frame_2 = IAM.Frame(...)
-frames = [frame_1, frame_2]
-```
-
-In other words, this representation captures that `frame_1` comes before `frame_2`, but does not express how much time has passed between the two frames. In order to express how much time has passed between the frames we need to specify the field `relative_timestamp` for each frame object. If we for example know that we have collected and aggregated our sensor data at 2Hz, then we would express that as
-
-```python
-frame_1 = IAM.Frame(..., relative_timestamp=0)
-frame_2 = IAM.Frame(..., relative_timestamp=500)
-frames = [frame_1, frame_2]
-```
-
-:::tip Why is relative time important?
-Specifying the time relationship between frames is important in order to enable different types of advanced annotator assistance tools in the Annotell platform. These tools enable a **significant** reduction in annotation time, while keeping quality high.
-:::
-
-## Creating the input
-In addition to supplying the sequential and temporal information for our `Frame` objects we also need to specify the camera images that constitute each frame. This is done by passing a list of `ImageFrame` objects, where each of these objects contains the path to the underlying file as well as the sensor name. Finally, we also need to specify the `frame_id` of each frame.
-
-In order to create the input we need to use our list of `Frame` objects and specify the parameters `external_id` and `sensor_specification`.
-
-| Parameter            | Description                                                                                                                                                                                                                                                      |
-| -------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| external_id          | Id which can be used to track progress of annotations with.                                                                                                                                                                                                      |
-| sensor_specification | Additional information about sensors, includes `sensor_to_pretty_name` and `sensor_order`. Defines which sensor that should be shown first, the sensor_order, or a mapping of sensor names to a prettier name version displayed in the Annotell Annotation Tool. |
 
 ```python reference
 https://github.com/annotell/annotell-python/blob/master/annotell-input-api/examples/cameras_seq.py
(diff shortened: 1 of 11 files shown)


docs-src/docs/input-api/overview.md
```diff
@@ -0,0 +1,228 @@
...
+```python
+frame_1 = Frame(..., relative_timestamp=0)
+frame_2 = Frame(..., relative_timestamp=500)
+frame_3 = Frame(..., relative_timestamp=1000)
+frames = [frame_1, frame_2, frame_3]
+```
+
+The `frame_id` is expressed as a string and is used to produce a unique identifier for each frame in the list of frames. The `frame_id` is used as a top-level key in the produced annotations, indicating which parts of the complete annotation belong to this specific frame.
+
+A common use case is to have the `frame_id` correspond to the `relative_timestamp` for each frame.
```
fredrikjacobson, line 164: Do we want to promote this as best practice?
Lets be **bold** and promote what we think is good. I've seen also seen uuids used which I think is preferable or external_id + frame_index

docs-src/docs/input-api/usage.md
```diff
@@ -147,3 +147,13 @@ client.annotation.get_annotations(input_uuids=[
...
 ```
+
+A common use case is downloading all completed annotations a project. This can be performed by first fetching all of the inputs added to the project and then filtering to only include inputs with status `"created"` (see [Input Statuses](#input-status)).
+
+```python
+project = "project-identifier"
+inputs = client.input.get_inputs(project=project)
+input_ids = [input.uuid for input in inputs if input.status == "created"]
+annotations = input_api.annotation.get_annotations(input_uuids=input_ids)
+
```
fredrikjacobson, line 158: would it be possible to have it as part of our tests and link via github? 
Seems like we want to ensure the integrity of this one. Could make use of e.g. `cameras-project`

fredrikjacobson (approved):
I like where this is heading. 

small suggestions on how to ensure our example code works.

docs-src/docs/input-api/overview.md
cotramarko: Sure, I can change the text to say "best practice is to use uuids or a combination of external_id + frame_index". But before that, @Ragge-dev do you have any feedback on this as well? Want to check that we all share the same view of what's recommended here :)

docs-src/docs/input-api/usage.md
cotramarko: Hmm that's a good point. I can add a small test for that on master and then include it in the docs

cotramarko: Done

docs-src/docs/input-api/overview.md
Ragge-dev: I think both seems ok:) Either you have very unique key or you have the ordering of the frames encoded in the keys, think both should be allowed

docs-src/docs/input-api/overview.md
```diff
@@ -0,0 +1,228 @@
...
+Sequential inputs are similarly represented, except that they instead contain a list of Frames
+
+```python
+@dataclass
+class InputSeq():
+    external_id: str
+    frames: List[Frame]
+    sensor_specification: SensorSpecification
+    calibration_id: Optional[str] # Required if using lidar sensors
+```
```
Ragge-dev, line 44: We should have startTs here as well, or is the plan that whoever adds it in the python models also adds it here?

docs-src/docs/input-api/overview.md
```diff
@@ -0,0 +1,228 @@
...
+
+The fields contain all of the information required to create the input.
+
+### External Id
+Whenever an input is uploaded it automatically gets an unique UUID, this is used as the primary identifier by Annotell and by all of our internal systems. However, in order to make communication around specific inputs easier we also allow for clients to include any kind of identifier to the input via the external id.
```
Ragge-dev, line 49: unique is already in the name (Universally Unique Identifier) so it reads weirdly to me.
```suggestion
Whenever an input is uploaded it automatically gets an UUID, this is used as the primary identifier by Annotell and by all of our internal systems. However, in order to make communication around specific inputs easier we also allow for clients to include any kind of identifier to the input via the external id.
```

docs-src/docs/input-api/overview.md
Ragge-dev: I feel like I answered my own question :sunglasses:

cotramarko: Since we don't use it for anything at the moment I think it's best to skip it entirely :P

cotramarko: As such, what do you @Ragge-dev @fredrikjacobson think about removing start_ts entirely from the python-model? Then we can introduce it in the python model once we actually start using it for something

docs-src/docs/input-api/inputs/cameras_seq.md
```diff
@@ -2,39 +2,8 @@
...
-| -------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| external_id          | Id which can be used to track progress of annotations with.                                                                                                                                                                                                      |
-| sensor_specification | Additional information about sensors, includes `sensor_to_pretty_name` and `sensor_order`. Defines which sensor that should be shown first, the sensor_order, or a mapping of sensor names to a prettier name version displayed in the Annotell Annotation Tool. |
 
 ```python reference
 https://github.com/annotell/annotell-python/blob/master/annotell-input-api/examples/cameras_seq.py
```
Ragge-dev, line 9: I renamed these files in order to add an example with videos as well, so this link is broken atm (doesn't load in the docs)
```suggestion
https://github.com/annotell/annotell-python/blob/master/annotell-input-api/examples/cameras_seq_images.py
```

docs-src/docs/input-api/usage.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: General
+title: Working with Inputs
```
Ragge-dev, line 2: Line 8, we are missing Cameras_seq inputs, not a line you have changed but still. However I would suggest that we simply refer people to the input-types section, so that we don't need to update this list when we add new inputs

docs-src/docs/input-api/overview.md
Ragge-dev: It sounds good to me, then we can think about moving it to the metadata as well without breaking any interface.

Ragge-dev (approved):
Excellent work:) Fix the broken link in cameras_seq and I am ok with a merge. I think this greatly improves on our current docs

PR #59, opened 2021-04-08
make _client and _file_client available

fredrikjacobson:
Allows easier subclassing for extended clients


cotramarko (approved):
LGTM :shipit:

fredrikjacobson:
@cotramarko I'll do the same for our resources, that way intellisense will not suggest the use of client when doing
`client.project...`

Ragge-dev (approved):
Good thinking!

PR #60, opened 2021-04-09
Remove start_timestamp

cotramarko:


fredrikjacobson:
Why are we removing it from the model?

Ragge-dev:
> Why are we removing it from the model?

@cotramarko is it only removed from cameras_seq or from all inputs?

@fredrikjacobson I want to remember that we talked about that 
1. It is a bit strange that it is a top field and 
2. It is not currently used anywhere in the codebase, neither frontend nor any ml-apps

And I remember at least me and Marko discussing in Slack whether we really need it in the client right now or if it is just another cause of confusion. I can't find exactly where though, it might be worth to discuss this again before merging

cotramarko:
@Ragge-dev  It's removed from all inputs, as in it was only present on cameras_seq in the python code before this PR :P

PR #61, opened 2021-04-09
Remove unused frame_id & relative_timestamp parameters

cotramarko:


annotell-input-api/annotell/input_api/model/input/lidars_and_cameras/frame.py
```diff
@@ -2,19 +2,15 @@
 from dataclasses import dataclass, field
 from annotell.input_api.model.input.resources.image import Image
 from annotell.input_api.model.input.resources.point_cloud import PointCloud
-
+from annotell.input_api.model.input.abstract import *
```
fredrikjacobson, line 5: @Ragge-dev did we reach a conclusion regarding the wildcard imports?

annotell-input-api/annotell/input_api/model/input/lidars_and_cameras/frame.py
```diff
@@ -2,19 +2,15 @@
...
 
 @dataclass
-class Frame:
-    frame_id: str
-    relative_timestamp: int
+class Frame(CameraFrame):
     point_clouds: List[PointCloud] = field(default_factory=list)
     images: List[Image] = field(default_factory=list)
```
fredrikjacobson, line 10: I'm not sure about this subclassing, because video_frame will appear in the argument list now. 
image

Do we still require this subclassing as part of setting the image dimensions?

fredrikjacobson (commented):
I guess this is a face where dataclasses, inheritance, and interfaces are somewhat mixed together in python. While we want the frames to implement the CameraFrame interface, we don't want them to expose `videos_frames` in the constructor, but rather as a static field. 

One way to go about it is to inherit `CameraFrame` but not use a dataclass, then implement a constructor taking only the fields exposed in the class here.

annotell-input-api/annotell/input_api/model/input/lidars_and_cameras/frame.py
cotramarko: This is the only method that relies on the CameraFrame class: 
```
    @staticmethod
    def _get_camera_dimensions(camera_frame: IAM.CameraFrame) -> Dict[str, IAM.CameraSettings]:
        def _get_camera_settings(width_height_dict: dict) -> IAM.CameraSettings:
            return IAM.CameraSettings(width_height_dict['width'], width_height_dict['height'])

        image_settings = {}
        video_settings = {}
        if len(camera_frame.images) != 0:
            image_settings = {
                image.sensor_name:
                    _get_camera_settings(get_image_dimensions(image.filename)) for image in camera_frame.images
             }
        if len(camera_frame.video_frames) != 0:
            # TODO: Support this for VIDEO
            pass

        return {**image_settings, **video_settings}
```

Since we don't do anything with video_frames we might just as well skip it entirely from the CameraFrame class?

annotell-input-api/annotell/input_api/model/input/lidars_and_cameras/frame.py
Ragge-dev: From what I have experienced it is ok to use wildcars if you want to import an entire package (So this case is fine).
However we should avoid it when importing classes from a module we should **not** use wildcards. Then we can import much more than just the classes we have defined in the module

annotell-input-api/annotell/input_api/model/input/lidars_and_cameras/frame.py
Ragge-dev: My preferred solution would be to not have this class inherit CameraFrame since it does not include video_frame:s, and there are other classes that inherit CameraFrame and uses video_frame:s (e.g. camera_sequence).

As you say @cotramarko that's the only method that uses the class, so when I finish my current work that method will dissapear since we won't set the dimensions in python and then I will remove CameraFrame as well. So I think it's unnecessary to make any change to the that class, since it will soon dissapear

cotramarko: Okay, in that case I will merge this PR and then in your PR you will omit CameraFrame entirely

PR #63, opened 2021-04-14
Add metadata support, both on an input-level but also on a frame-level for sequence inputs

cotramarko:


annotell-input-api/annotell/input_api/model/input/lidars_and_cameras/lidars_and_cameras.py
```diff
@@ -15,12 +15,14 @@ class LidarsAndCameras(CameraInput):
     frame: Frame
     calibration_id: str
     sensor_specification: SensorSpecification
+    metadata: Mapping[str, Any] = field(default_factory=dict)
```
fredrikjacobson, line 18: should we type this as `Map[str, Union[str, int, float]`? 
Sort of want to discourage nested JSON

fredrikjacobson (commented):
one question wr.t. type hints

fredrikjacobson:
Also change me :)

annotell-input-api/annotell/input_api/model/input/lidars_and_cameras/lidars_and_cameras.py
cotramarko: Hmm, I was thinking a bit about that. Either we force it as those three, or we allow whatever (as long as it's serializable) but we will only treat it as a KV-pair, and not do any processing of nested structures.

The only difference from a client-perspective is that you can either do `a = {"key": value_dict}` or `a = {"key": str(value_dict)}`

cotramarko:
> Also change notes

Will fix that, good that you brought it up

cotramarko:
Updated type-hints, changed examples to include proper metadata as well

Ragge-dev (commented):
What is the argument for not having metadata as Optional in both the input and frame objects (in this client that is)? To me this would be the natural approach.

cotramarko:
Personal preference I guess 🤷‍♂️  I'm not a big fan of optional iterables, since what's the difference between an empty iterable and None? 

Additionally, since we on the backend store an empty mapping in the DB as the default behaviour (for backwards compatibility) it made more sense to me to have consistent behaviour on the client-side as well -- i.e. if you don't want to include any explicit metadata it will instead default to an empty mapping, and this is what you get back from the system as well (instead of having `metadata=None` but when retrieving the associated entity you see that it has `metadata={}`)

Ragge-dev:
The difference to me would be that if I don't want to include any metadata I would just ignore it instead of having to write the extra boiler-plate "metadata=dict()" when I create all my objects. It feels weird to me that this isn't an opt-in feature instead of having everyone needing to specify their metadata, even if there is none.

PR #64, opened 2021-04-19
Removed sensor_settings logic

Ragge-dev:
Solves the python part of https://annotell.atlassian.net/browse/INT-179

Removed all logic related to sensor_settings in the python client


fredrikjacobson (approved):
Great. 

Thats a lot of code for supplying pixel dimensions.

annotell-input-api/annotell/input_api/model/input/cameras/cameras.py
```diff
@@ -1,20 +1,17 @@
...
+class Cameras:
     external_id: str
-    sensor_specification: SensorSpecification
     frame: Frame
+    sensor_specification: Optional[SensorSpecification] = None
 
     def to_dict(self) -> dict:
         return dict(frame=self.frame.to_dict(),
-                    sensorSpecification=self.sensor_specification.to_dict(),
+                    sensorSpecification=self.sensor_specification.to_dict() if self.sensor_specification is not None else None,
```
cotramarko, line 16: What do you think about writing it like this instead? :) 
```
... if self.sensor_specification else None
```

cotramarko (approved):
LGTM

annotell-input-api/annotell/input_api/model/input/cameras/cameras.py
Ragge-dev: I think I like your way better actually, my default is to always write like this because I want to be defensive about the False case (like allowing empty strings, arrays etc), but my goal was here actually to be defensive about the True case.

With that said, what do you think about writing it something like this?
`... if isinstance(self.sensor_specification, SensorSpecification) else None`

cotramarko: That's even better!

Ragge-dev: Fixed

PR #66, opened 2021-04-26
Int 271 input view links

Ragge-dev:
python part of https://annotell.atlassian.net/browse/INT-271

Added view_link to Input model and changed annoutil to use the url in the input


cotramarko (approved):
LGTM, good work :)

PR #71, opened 2021-05-10
Int 311 resource id docs

cotramarko:
This is the rendered version: 

image

### DIFF
--- a/docs-src/docs/input-api/overview.md
+++ b/docs-src/docs/input-api/overview.md
@@ -235,3 +235,46 @@ Notice how we for each video frame need to specify the timestamp in the video fi
 The choice of which approach to use is up to the client. Long sequences should preferrably be encoded as videos, since this will result in smaller amounts of egress required, resulting in annotations being produced faster. However, this requires care with respect to image quality since most video encodings can degrade quality depending on the chosen compression, codec and bitrate.
 
 If image quality if of central importance and sequnces are not large then it's recommended to supply individual camera images instead of videos.
+
+## Image, Video & Pointcloud Resources
+Every single file containing binary sensor data (e.g. image, video or pointcloud files) is represented as a `Resource`.
+
+```python reference
+https://github.com/annotell/annotell-python/blob/f2b941373b1dff4297d7705ef0f2587eadbca7b3/annotell-input-api/annotell/input_api/model/input/resources/resource.py#L7-L12
+```
+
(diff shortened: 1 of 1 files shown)


docs-src/docs/input-api/overview.md
```diff
@@ -235,3 +235,42 @@ Notice how we for each video frame need to specify the timestamp in the video fi
...
+Every single file containing binary sensor data (e.g. image, video or pointcloud files) is represented as a `Resource`, with `Image`, `PointCloud` and `VideoFrame` all being subclasses of it. 
+
+```python reference
+https://github.com/annotell/annotell-python/blob/f2b941373b1dff4297d7705ef0f2587eadbca7b3/annotell-input-api/annotell/input_api/model/input/resources/resource.py#L7-L12
+```
+
+When specifying a `Resource` object (like `Image`, `PointCloud` or `VideoFrame`) it's possible to either:
+
+1. Refer to _local_ files, these will be uploaded (synchronously) to the Annotell platform.
+2. Refer to _remote_ files via URI, these will only be uploaded (asynchronously) and stored in the Annotell platform if necessary. Otherwise they will be served to annotators via the URI.
```
Ragge-dev, line 249: Could we make it a bit more clear that it is only necessary if special processing is needed of the data, e.g. converting to potree? I ask myself here what's "necessary"

docs-src/docs/input-api/overview.md
```diff
@@ -235,3 +235,42 @@ Notice how we for each video frame need to specify the timestamp in the video fi
...
+
+**Alternative 2** is achieved by setting the parameter `filename` to just be the filename, and setting the parameter `resource_id` to the corresponding URI of the file, e.g. 
+
+```python
+Image(
+    "vid_FC.mp4",
+    sensor_name='FC',
+    resource_id="gs://data-collection/4fcc30af/vid_FC.mp4"
+)
+```
```
Ragge-dev, line 270: In texts describing alternative 1 and 2 we mention the parameter `filename` but it is not in the example code, I think it would be good to include it there as well

Ragge-dev (commented):
Great work Marko :+1: I had two small things that I suggested to make the text a bit more clear

docs-src/docs/input-api/overview.md
cotramarko: Great point :)

docs-src/docs/input-api/overview.md
Ragge-dev: Works for me:)

PR #73, opened 2021-05-19
Int 330 calibration

fredrikjacobson:
Adds calibration when fetching inputs

### DIFF
--- a/annotell-input-api/README.md
+++ b/annotell-input-api/README.md
@@ -13,7 +13,7 @@ All notable changes to this project will be documented in this file.
 
 ## [1.0.6] - Not published
 ### Added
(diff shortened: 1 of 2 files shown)


annotell-input-api/README.md
```diff
@@ -11,6 +11,10 @@ Documentation about how to use the library can found [here](https://annotell.git
 
 All notable changes to this project will be documented in this file.
 
+## [1.0.6] - Not published
+### Added
+- `calibration_id` not available for created inputs via the `client.input.get_inputs` method.
```
cotramarko, line 16: Wait a minute, this should say "now" instead of "not", right?

PR #75, opened 2021-05-26
Annotation types in docs

Ragge-dev:
Added in the docs a section about Annotation Types, about creating batches and fixed some errors


fredrikjacobson (approved):
Just add a snippet on publish previous and we're good to go

PR #76, opened 2021-05-28
Add publish option to create batch

fredrikjacobson:


Ragge-dev (approved):
Looking forward to see how clients will like this

PR #79, opened 2021-06-17
Int 352 inputs with UUID

Ragge-dev:
https://annotell.atlassian.net/browse/INT-406

It was strange to me that this was not available in the python client, since we return the uuid from the create method. And I also needed this for another test I was writing


Ragge-dev:
Didn't do this from master, will redo

cotramarko (commented):
Nice!

PR #81, opened 2021-06-17
Added input_uuids as parameter to get_inputs method

Ragge-dev:
Solves https://annotell.atlassian.net/browse/INT-406

Thought it was strange that get_inputs method doesn't take the input_uuid since we respond with it from the create methods. I also needed this for a test I'm writing

### DIFF
--- a/annotell-input-api/annotell/input_api/resources/input/input.py
+++ b/annotell-input-api/annotell/input_api/resources/input/input.py
@@ -32,7 +32,8 @@ class InputResource(InputAPIResource):
         project: str,
         batch: Optional[str] = None,
         include_invalidated: bool = False,
-        external_ids: Optional[List[str]] = None
+        external_ids: Optional[List[str]] = None,
+        input_uuids: Optional[List[str]] = None,
     ) -> List[InputModel.Input]:
         """
         Gets inputs for project, with option to filter for invalidated inputs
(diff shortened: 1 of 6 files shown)


annotell-input-api/annotell/input_api/resources/input/input.py
```diff
@@ -41,14 +42,17 @@ def get_inputs(
         :param batch: Batch (identifier) to filter
         :param invalidated: Returns invalidated inputs if True, otherwise valid inputs
         :param external_id: External ID to filter input on
+        : param input_uuids: A UUID to filter inputs on
         :return List: List of Inputs
         """
 
         external_id_query_param = ",".join(external_ids) if external_ids else None
+        input_uuids_query_param = ",".join(input_uuids) if input_uuids else None
```
fredrikjacobson, line 50: Should we bite the bullet and change it into a `POST /inputs/query`?
I suppose we can ship this but we should switch to supplying in the body, preferably with a POST then I suppose.

fredrikjacobson (approved):
I think it's OK to merge since we expose this from our API:s.
Only thing I worry about is the request URI limit.

annotell-input-api/annotell/input_api/resources/input/input.py
Ragge-dev: Completely agree, should do the same with download annotations while we are at it. I'll skip it in this PR but create a JIRA

cotramarko: Yep, good point

cotramarko (approved):

PR #86, opened 2021-08-05
Added method to get inputs using only uuids

Ragge-dev:
Solves: https://annotell.atlassian.net/browse/INT-406

### DIFF
--- a/annotell-input-api/README.md
+++ b/annotell-input-api/README.md
@@ -13,7 +13,8 @@ All notable changes to this project will be documented in this file.
 
 ## [1.0.8] - TBD
 ### Added
-The `get_inputs` method now accepts `input_uuids` as a parameter.
+- A new method has been added, `get_inputs_with_uuids`, which can fetch inputs using only the `input_uuid`.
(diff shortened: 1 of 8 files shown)


cotramarko (commented):
Nice!

However, shouldn't this also be added to `annoutil`?

Ragge-dev:
> Nice!
> 
> However, shouldn't this also be added to `annoutil`?

Added it to annoutil.
I am all for removing that though, in favor of doing something in the app.

fredrikjacobson:
> > Nice!
> > However, shouldn't this also be added to `annoutil`?
> 
> Added it to annoutil.
> I am all for removing that though, in favor of doing something in the app.

What does doing something in the app mean?

Ragge-dev:
> > > Nice!
> > > However, shouldn't this also be added to `annoutil`?
> > 
> > 
> > Added it to annoutil.
> > I am all for removing that though, in favor of doing something in the app.
> 
> What does doing something in the app mean?

I was refering to our discussion of removing annoutil alltogether, it has nothing to do with this PR it was just on the top of my mind... Lets take that outside of this PR.

PR #89, opened 2021-09-10
Changed name of OpenLabelData to OpenLabelAnnotation and also changed so that imports will be simpler

olinoskar:


alydh-annotell (approved):

alydh-annotell:
Please update in the main annotell repo as well

PR #91, opened 2021-09-13
add download annotation by annotation type

fredrikjacobson:
Add support for fetching annotations via project/batch/annotation-type. 

`get_annotations` will return a `Generator[dict]`, meaning that we will yield the annotations and not fetch them in one request. I expect this to be **slow**, but it gives us the flexibility to change the implementation of internal storage.

This way you can either iterate or call `list` on the generator. (Which we need to add a notice on in the docs). 

Further, I expect `get_annotation` to be the primary way of downloading annotations when using callbacks.

### DIFF
--- a/annotell-input-api/annotell/input_api/model/annotation/client_annotation.py
+++ b/annotell-input-api/annotell/input_api/model/annotation/client_annotation.py
@@ -10,7 +10,7 @@ class Annotation(Response):
     input_uuid: str
     annotation_type: str
     created: datetime
-    annotation: Optional[Dict] = None
+    content: Dict
 
     @staticmethod
     def from_json(js: dict):
@@ -18,5 +18,5 @@ class Annotation(Response):
             input_uuid=js["inputUuid"],
             annotation_type=js["annotationType"],
             created=ts_to_dt(js["created"]),
-            annotation=js.get("annotation")
+            content=js.get("content") or dict()
         )
--- a/annotell-input-api/annotell/input_api/model/annotation/client_annotation.py
+++ b/annotell-input-api/annotell/input_api/model/annotation/client_annotation.py
@@ -1,0 +1,22 @@
+from typing import Optional, Dict
+from datetime import datetime
+from dataclasses import dataclass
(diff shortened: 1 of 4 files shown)


annotell-input-api/annotell/input_api/resources/annotation/annotation.py
```diff
@@ -1,33 +1,53 @@
...
         return annotations
+
+    def get_project_annotations(self,
+                                project: str,
+                                annotation_type: str,
+                                batch: Optional[str] = None) -> Generator[dict, None, None]:
+        url = f"v1/annotations/projects/{project}/"
+        if batch:
+            url += f"batch/{batch}/"
+        if annotation_type:
```
olinoskar, line 41: Shouldn't we always include `annotation_type` in the url?

olinoskar (changes requested):

annotell-input-api/annotell/input_api/resources/annotation/annotation.py
fredrikjacobson: Yes!

olinoskar (approved):

annotell-input-api/annotell/input_api/model/annotation/client_annotation.py
```diff
@@ -0,0 +1,22 @@
...
+from annotell.input_api.model.abstract.abstract_models import Response
+from annotell.input_api.util import ts_to_dt
+
+
+@dataclass
+class Annotation(Response):
+    input_uuid: str
+    annotation_type: str
+    created: datetime
+    annotation: Optional[Dict] = None
```
cotramarko, line 13: Should this be `Optional`?

cotramarko (changes requested):

annotell-input-api/annotell/input_api/model/annotation/client_annotation.py
fredrikjacobson: I had ideas about exposing the list entity itself, but might actually change to only exposing the resolved one, which will remove this optionality, and expose `Annotation`  which will be a class

PR #103, opened 2021-11-02
1.0.X download_annotations is not yet deprecated!

Ragge-dev:
In the docs it currently says that the annotation models/method related to v.1.0.X are deprecated, but they are not yet! But they will be, might be confusing so I wanted to change it back!


alydh-annotell (approved):

alydh-annotell:
:football: :goal_net:

PR #105, opened 2021-11-09
Int 593 update docs calibration

alydh-annotell:


docs-src/docs/input-api/calibration.md
```diff
@@ -1,45 +1,97 @@
...
+
 ```python reference
-https://github.com/annotell/annotell-python/blob/master/annotell-input-api/examples/calibration.py
+https://github.com/annotell/annotell-python/blob/master/annotell-input-api/examples/calibration/calibration.py
 ```
 
-:::tip reuse calibration
-Note that you can, and should, reuse the same calibration for multiple inputs if possible.
+:::tip reuse calibration Note that you can, and should, reuse the same calibration for multiple inputs if possible.
 :::
```
fredrikjacobson, line 85: I think this will render a bit weirdly.

fredrikjacobson (approved):

PR #107, opened 2021-11-18
Int 757 update openlabel schema

alydh-annotell:


olinoskar:
I think that we should fix ontologies, resources and resource uids before merging this PR

alydh-annotell:
> I think that we should fix ontologies, resources and resource uids before merging this PR

Could you suggest how to fix it here in the PR or make a commit? Not sure what changes you mean

olinoskar (approved):

PR #108, opened 2021-11-22
Docs v11x openlabel

olinoskar:
Updated documentation to version 1.1.x and added documentation about OpenLabel. Would appreciate som extra feedback on:
- The examples
- OpenLabel format


docs-src/docs/openlabel/openlabel-format.md
```diff
@@ -0,0 +1,144 @@
...
+that the format is valid.
+
+Even though OpenLABEL is a strict format there is still some room for interpretation. In this section we try to clarify
+some of these parts and explain the choices that we have made within the standard.
+
+## Non-sequences are sequences with one frame
+
+Due to reasons of simplicity we have made the choice to treat non-sequences in the same way as sequences. This
+means that non-sequences are represented as a sequence with only one frame. Only data such as `name` and `type` are defined
+in the top level element keys. All other information is
```
Ragge-dev, line 17: sentence abruptly ends! :eyes:

olinoskar: fixed

docs-src/docs/openlabel/openlabel-format.md
```diff
@@ -0,0 +1,144 @@
...
+        "name": "0",
+        "type": "isFollowing",
+        "rdf_subjects": [{"type": "object", "uid": "1"}],
+        "rdf_objects": [{"type": "object", "uid": "0"}]
+      },
+      "1": {
+        "name": "1",
+        "type": "isFollowing",
+        "rdf_subjects": [{"type": "object", "uid": "Unclear"}],
+        "rdf_objects": [{"type": "object", "uid": "2"}]
```
Ragge-dev, line 97: Is this correct, I thought that the if car-0 is following car-1 then car-0 would be the subject and car-1 the object?

docs-src/docs/key_concepts.md
```diff
@@ -79,7 +79,25 @@ Inputs can be created via Annotell's Input API, which has support for several di
...
+called `annotell-openlabel` which you can find [here](openlabel/python-client). With this
+it is straight forward to parse the json. Here is an example where the annotation is fetched using a method described in
+the [Downloading Annotations](input-api/annotations.md) chapter:
+```
+client = InputApiClient()
+annotation: Annotation = client.annotation.get_annotation(input_uuid=input_uuid, annotation_type=annotation_type)
+open_label: OpenLabelAnnotation = OpenLabelAnnotation.parse_obj(annotation.content)
+```
+
+Any conversion away from this format will need to occur client-side.
```
Ragge-dev, line 97: I'm thinking if we can remove this and instead just reference to the other sections that we have in the docs, I think with all the good examples you have created this is no longer necessary

docs-src/docs/openlabel/openlabel-format.md
```diff
@@ -0,0 +1,144 @@
+---
+title: OpenLABEL format
```
Ragge-dev, line 2: I think this section turned out great, really well explained I think and will help with FAQ about out OpenLABEL export format

docs-src/docs/input-api/annotations.md
```diff
@@ -56,4 +35,70 @@ This method returns a single `Annotation` object, containing the OpenLABEL json,
 https://github.com/annotell/annotell-python/blob/master/annotell-input-api/examples/get_project_annotations.py#L8-L23
 ```
 This example fetches annotations for an entire project or batch. The `run()` method returns a generator which will yield `Annotation` objects for all finished annotations
-, for the given `project`, `batch` and `annotation_type`, and in the end prints all of these annotations.
\ No newline at end of file
+, for the given `project`, `batch` and `annotation_type`, and in the end prints all of these annotations.
+
+
+## Examples
```
Ragge-dev, line 41: Since what we have above are examples as well, maybe this section could be named something like "common use cases"?

Ragge-dev (commented):
Great work, I think this will be a **great** addition to our docs:)

docs-src/docs/openlabel/openlabel-format.md
```diff
@@ -0,0 +1,144 @@
...
+      }
+    ]
+  }
+}
+```
+
+
+## Relations
+
+We consider two types of relation; one-directional relation between two objects and group relations.
```
cotramarko, line 69: Really picky, but consider replacing "one-directional" with "unidirectional" instead 🤓

docs-src/docs/openlabel/openlabel-format.md
cotramarko: I agree with Ragnar, this is really awesome! ⭐

cotramarko:
Great stuff! 

One thought is if we should perhaps try to move some of the examples into separate `.py`-files that we reference to with `python reference`? For one, it's a bit easier to version control it this way. Additionally we could also add tests to the code examples so that we have some safety in the future as we change and update a lot of the code.

docs-src/docs/openlabel/openlabel-format.md
olinoskar: You're right. Fixing this

olinoskar: I think you're right. Unfortunately I think this is wrong in the openlabel conversion though. Fixing the docs here and creating a bug report for the converter

docs-src/docs/key_concepts.md
olinoskar: hmm, isn't it nice to have that there too though? It is one of out key concepts

docs-src/docs/input-api/annotations.md
olinoskar: Sounds better. Changing to this

olinoskar:
> Great stuff!
> 
> One thought is if we should perhaps try to move some of the examples into separate `.py`-files that we reference to with `python reference`? For one, it's a bit easier to version control it this way. Additionally we could also add tests to the code examples so that we have some safety in the future as we change and update a lot of the code.

@Ragge-dev and I were discussing this yesterday ended it ended with the sentiment "let's think about it". We see the same advantages as you do with moving some of them into separate python files. So unless @Ragge-dev has any objections I think we should do this.

cotramarko:
> > Great stuff!
> > One thought is if we should perhaps try to move some of the examples into separate `.py`-files that we reference to with `python reference`? For one, it's a bit easier to version control it this way. Additionally we could also add tests to the code examples so that we have some safety in the future as we change and update a lot of the code.
> 
> @Ragge-dev and I were discussing this yesterday and it ended with the sentiment "let's think about it". We see the same advantages as you do with moving some of them into separate python files. However one of the drawbacks of doing that is that `annotell-input-api` would become dependent on `annotell-openlabel`, which I don't know if we want

Aha, I see. I leave it to you to decide on this, nice that you already had it in mind :)

docs-src/docs/openlabel/openlabel-format.md
```diff
@@ -0,0 +1,197 @@
...
+            "type": "isSubjectOfAction",
+            "rdf_subjects": [{"type": "object", "uid": "1"}],
+            "rdf_objects": [{"type": "action", "uid": "1"}]
+        }
+    },
+    "actions": {
+        "0": {"name": "", "type": "Road"},
+        "1": {"name": "Unclear", "type": "Road"}
+    }
+}
```
Ragge-dev, line 137: In the jira that you create for moving the python code to the input-api repo I think it would be good if we at the same time move these examples to the annotell-openlabel repo:) It will be easier then as well to keep track of these examples

Ragge-dev (approved):

PR #109, opened 2021-11-24
Fixed so stream properties are generated properly

alydh-annotell:


olinoskar (commented):
Nice work! The only thing I am thinking about is that is might be nice to have separate definitions for the fisheye and pinhole models under stream properties. This would give them more informative names (rather than `StreamProperty` and `StreamProperty1`) and I think it would make it clearer both in the schema and in the models.

alydh-annotell:
> 

I agree, but at this instance we are not supporting including the calibrations in openlabel, so I think we can change that when/if implement that

Ragge-dev:
> > Nice work! The only thing I am thinking about is that is might be nice to have separate definitions for the fisheye and pinhole models under stream properties. This would give them more informative names (rather than StreamProperty and StreamProperty1) and I think it would make it clearer both in the schema and in the models.
> 
> I agree, but at this instance we are not supporting including the calibrations in openlabel, so I think we can change that when/if implement that

But if the classes models fisheye and pinhole calibrations, and the names are terrible, why not change the names now instead? Keeping it will just end up in hard to read code

annotell-openlabel/annotell/openlabel/models/models.py
```diff
@@ -247,9 +235,36 @@ class Config:
     width_px: Optional[int] = None
 
 
-class StreamProperties(BaseModel):
-    __root__: Union[Dict[str, Any], StreamProperty, StreamProperty1] = Field(
-        ..., description='Properties of the stream.'
+class Type4(Enum):
```
Ragge-dev, line 238: Could we call this StreamType instead?

PR #114, opened 2021-12-28
Updated models for better type casting

olinoskar:
Jira: [INT-802](https://annotell.atlassian.net/jira/software/c/projects/INT/boards/41?modal=detail&selectedIssue=INT-802)

Please have an extra look at `custom_types.py` and the changes made to the `Stream` model.


annotell-openlabel/annotell/openlabel/models/models.py
```diff
@@ -469,7 +469,7 @@ class Config:
         None,
         description='Streams is a JSON object which contains OpenLABEL streams with specific information for this frame. Stream keys can be any string, for example, a friendly stream name.',
     )
-    timestamp: Optional[Union[str, float]] = Field(
+    timestamp: Optional[Timestamp] = Field(
```
Ragge-dev, line 472: I would really like for us to manage without custom types, lets discuss if that is possible. The reason is that if a new version of openlabel is released it is more work (I think at least) to find all places where we need a custom type. It should be less work to switch the order so that the most specific type comes first, as described here https://pydantic-docs.helpmanual.io/usage/types/#unions

annotell-openlabel/annotell/openlabel/models/models.py
```diff
@@ -1023,7 +1023,7 @@ class Config:
         None,
         description='This attribute specifies whether the number shall be considered as a value, a minimum, or a maximum in its context.',
     )
-    val: float = Field(..., description='The numerical value of the number.')
+    val: Number = Field(..., description='The numerical value of the number.')
```
Ragge-dev, line 1026: Any reason this cannot be a float?

Ragge-dev (changes requested):
LineReference still has its values as float, should be an integer since it is a list of indexes

annotell-openlabel/annotell/openlabel/models/models.py
```diff
@@ -217,7 +217,7 @@ class StreamProperty1(BaseModel):
     """
 
     class Config:
-        extra = Extra.allow
+        extra = Extra.forbid
```
Ragge-dev, line 220: Why switch to not allowing extra fields?

annotell-openlabel/annotell/openlabel/models/utils/custom_types.py
```diff
@@ -0,0 +1,31 @@
...
+                return int(v)
+            except ValueError:
+                try:
+                    return float(v)
+                except ValueError:
+                    return v
+
+        if isinstance(v, float):
+            return v
+        return v
```
alydh-annotell, line 20: Even if we keep this code in some sense, here are some suggestions to make it more concise.

```suggestion
        if not isinstance(v, str):
            return v

        if v.isdigit():
            return int(v)

        return float(v)
```

Idk how the validation works internally for Pydantic with validators, but don't we have to raise an exception if the validation fails somehow?

In my suggestion the only failure-mode is if the cast to float fails, which seems fine to me.

PR #117, opened 2022-01-05
This Ain't Scala 

alydh-annotell:


Ragge-dev (commented):
:c_shame: Thank you and well done:) 

Before merging, could you update README.md and the version in the __init__ file? I'm thinking that  want to publish a new version of the client right away

PR #122, opened 2022-01-31
WIP: Gh pages imu data csv fix

alydh-annotell:


docs-src/docs/input-api/resources/pointclouds.md
```diff
@@ -23,7 +23,7 @@ DATA ascii
 We validate `VERSION`, `FIELDS`, `SIZE`, `TYPE` and `COUNT` rows of this header. All other formats will result in an error and the conversion will fail.
 
 ## CSV
-We currently only support the following exact header and separation character (where `intensity` is uint8 and `ts_gps`, `x`, `y`, `z` are all float32):
+We currently only support the following exact header and separation character (where `intensity` is uint8, `ts_gps` is long and `x`, `y`, `z` are all float32):
```
Ragge-dev, line 26: I think long is not that specific, it can be different things for different languages. Since an integer is what is used by clients atm I think we will need it to be a at least a uint64. (could be signed as well, but no need for an epoch time) In the future I think this would be better as a float as well

docs-src/docs/input-api/inputs/lidars_with_imu_data.md
```diff
@@ -0,0 +1,17 @@
...
+
+By including IMU data of the ego vehicle, motion compensation can be performed on the pointclouds.
+
+```python reference
+https://github.com/annotell/annotell-python/blob/master/annotell-input-api/examples/lidars_with_imu_data.py
+```
+
+```python reference
+https://github.com/annotell/annotell-python/blob/master/annotell-input-api/examples/imu_data/create_imu_data.py
+```
```
Ragge-dev, line 13: I don't know if we should have this example here, to me this is mainly useful simply as a utility tool for us (similar to the code we have to create a unity calibration). The first example I think should be sufficient

Ragge-dev (changes requested):
The new chapter is not visible in the documentation, in order to add it there you will need to add it under a suitable place in `docs-src/sidebars.js`. I'm thinking this chapter could be under the "Inputs" tab, I feel like we probably want to break out a lot under "Overview" in due time and give e.g. Calibrations a separate tab.

When in docs-src you can run: `npm run docusaurus start` in order to view how the documentation will look

alydh-annotell:
LGTM

PR #123, opened 2022-02-03
Fixed bug with ciruclar imports

Ragge-dev:
Fixes circular imports bug in client

### DIFF
--- a/annotell-input-api/annotell/input_api/model/input/lidars_and_cameras/frame.py
+++ b/annotell-input-api/annotell/input_api/model/input/lidars_and_cameras/frame.py
@@ -1,7 +1,7 @@
 from typing import List, Optional
 from dataclasses import dataclass, field
 
-from annotell.input_api.model import UnixTimestampNs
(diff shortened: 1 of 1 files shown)


annotell-input-api/annotell/input_api/model/input/lidars_and_cameras/frame.py
```diff
@@ -10,11 +10,11 @@
 class Frame:
     point_clouds: List[PointCloud] = field(default_factory=list)
     images: List[Image] = field(default_factory=list)
-    unix_timestamp: Optional[UnixTimestampNs] = None
+    unix_timestamp: Optional[float] = None
```
alydh-annotell, line 13: Shoulnd't this be UnixTimestampNs?

alydh-annotell (approved):
Fix comment then let's go

annotell-input-api/annotell/input_api/model/input/lidars_and_cameras/frame.py
Ragge-dev: It should....!

PR #126, opened 2022-02-17
Int 915 add reserved keys to metadata

alydh-annotell:
Implementation of metadata containing reserved keyword `region`

Can still use a `dict` so it is backwards compatible.

Would like some extra feedback on the docstring for the `MetaDataContainer` class, is it sufficiently clear?


annotell-input-api/README.md
```diff
@@ -11,6 +11,13 @@ Documentation about how to use the library can found [here](https://annotell.git
 # Changelog
 
 All notable changes to this project will be documented in this file.
+## [1.1.X] - 2022-02-17
+
+### Added
+
+- MetaDataContainer added with reserved keyword `region`. 
+import annotell.input_api.input_api_client as IAC
```
Ragge-dev, line 19: Tried but import does not work as expected in markdown!

annotell-input-api/annotell/input_api/model/input/metadata/metadata.py
```diff
@@ -0,0 +1,32 @@
...
+    Attributes:
+        region: A string indicating the region the data was collected in. 
+            If there are annotation instructions associated with a specific
+            region, talk to your contact at Annotell to sync what should be
+            specified in the region attribute.
+    """
+    region: Optional[str]
+
+    class Config:
+        extra = "allow"
```
Ragge-dev, line 23: We need to validate that the metadata is not nested, I guess a @validator thingy could be useful here

annotell-input-api/annotell/input_api/model/input/metadata/metadata.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+MetaData = Union[MetaDataMapping, MetaDataContainer]
```
Ragge-dev, line 26: Regarding the name, would it be crazy to name MetaDataContainer Metadata and MetaData something like AllowedMetadata, MetadataUnioin... or the like? My thinking is that if the pydantic class is what we want to go towards long term we could reserve the Metadata name for that one somehow

annotell-input-api/annotell/input_api/model/input/metadata/metadata.py
alydh-annotell: Closed: Will not fix

alydh-annotell: We do all validation server side

PR #130, opened 2022-03-23
Document shutter times

petern-annotell:
Adds documentation for shutter timings specified via Image metadata.


alydh-annotell (approved):

docs-src/docs/input-api/inputs/lidars_and_cameras_seq.md
```diff
@@ -31,3 +31,11 @@ https://github.com/annotell/annotell-python/blob/master/annotell-input-api/examp
 :::note Coordinate Systems
 Note that both `position` and `rotation` for ego vehicle pose are with respect to the *local* coordinate system.
 :::
+
+## Shutter timings
```
Ragge-dev, line 35: I think this has more to do with imu_data than with lidars_and_cameras_seq, so I think it should be in that section instead. No need to do it now though, we can do it if we add some more for multivision

PR #142, opened 2022-04-21
WIP: Int 1103 add pre annotations support in input api client

petern-annotell:
Here's the basis for creating pre-annotations as part of LACS creation. We do expect to separate input/scene creation later so there's questions about the exact API that are unsettled.

This needs the input-api stub changes for validation (INT-1136). A question I have is can this be tested by mocking away the HTTP layer?


annotell-input-api/annotell/input_api/resources/abstract/creatable_input_api_resource.py
```diff
@@ -30,6 +32,7 @@ def _post_input_request(
         batch: Optional[str],
         annotation_types: Optional[List[str]] = None,
         imu_data: Optional[List[IMUData]] = None,
+        pre_annotations: Optional[OpenLabelAnnotation] = None,
```
Ragge-dev, line 35: Something I notice is that we cannot specify for what annotation-types a pre-annotation should be created. In this case we would create the same pre-annotation for each annotation_type specified.

annotell-input-api/annotell/input_api/model/input/preannotations/validated_pre_annotations.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class ValidatedPreAnnotations(BaseSerializer):
+    resource_id: str
+    signed_url: Optional[str]
+    preannotations: str
+
+
+class ValidatePreAnnotationsRequest(BaseSerializer):
+    preannotations: str
```
Ragge-dev, line 13: I'm thinking this should be an openlabel

annotell-input-api/annotell/input_api/model/input/preannotations/validated_pre_annotations.py
```diff
@@ -0,0 +1,14 @@
+from typing import Optional
+
+from annotell.input_api.model.calibration.common import BaseSerializer
+
+
+class ValidatedPreAnnotations(BaseSerializer):
+    resource_id: str
+    signed_url: Optional[str]
+    preannotations: str
```
alydh-annotell, line 9: Is this supposed to be 1 pre-annotation or several?
If only one, please use singular names

annotell-input-api/examples/lidars_and_cameras_seq_with_pre_annotations.py
```diff
@@ -0,0 +1,112 @@
...
+    # Create calibration
+    calibration_spec = create_sensor_calibration(f"Collection {datetime.now()}", [lidar_sensor1, lidar_sensor2],
+                                                 [cam_sensor1, cam_sensor2])
+    created_calibration = client.calibration.create_calibration(calibration_spec)
+
+    # Generate IMU data
+    ONE_MILLISECOND = 1000000  # one millisecond, expressed in nanos
+    start_ts = 1648200140000000000
+    end_ts = start_ts + 10 * ONE_MILLISECOND
+    imu_data = create_dummy_imu_data(start_timestamp=start_ts, end_timestamp=end_ts, samples_per_sec=1000)
```
alydh-annotell, line 41: Think we can remove the imu_data and the image metadata in this example, to reduce the complexity

petern-annotell:
This is parked for the moment, do expect to respond to the feedback when it's picked up again.

annotell-input-api/annotell/input_api/resources/abstract/creatable_input_api_resource.py
```diff
@@ -110,6 +117,30 @@ def _upload_imu_data(self, validated_imu_data: Optional[ValidatedIMUData]) -> bo
             return True
         return False
 
+    def _validate_pre_annotations(self, body: Dict, preannotations: Optional[OpenLabelAnnotation], input_uuid: str,
+                                  dryrun: bool) -> Tuple[Dict, Optional[ValidatedPreAnnotations]]:
+        if None in (body, preannotations):
+            return body, None
+        preannotations_json = preannotations.json()
```
alydh-annotell, line 124:
```suggestion
        preannotations_json = preannotations.json(exclude_none=True)
```

alydh-annotell:
I changed this to be WIP.
Want to define the API side of things before implementing things in the client

PR #156, opened 2022-06-15
Int 1198 update the python client for lidar fov

petern-annotell:
Adds an optional FOV to LIDAR calibrations. This is captured as start/stop angle in degrees, and optional sensor depth (distance).


annotell-input-api/annotell/input_api/model/calibration/lidar/lidar_calibration.py
```diff
@@ -1,5 +1,15 @@
...
 
+class LidarFieldOfView(BaseSerializer):
+    start_angle_deg: float
+    stop_angle_deg: float
+    depth: Optional[float] = None
+
+
 class LidarCalibration(BaseCalibration):
     calibration_type = CalibrationType.LIDAR
+    field_of_view: Optional[LidarFieldOfView]
```
alydh-annotell, line 15: I think we should provide a default here
```suggestion
    field_of_view: Optional[LidarFieldOfView] = None
```

alydh-annotell (approved):
Approved after to yay or nay my suggestion

PR #161, opened 2022-06-23
Int 1205 update documentation for motion compensation

petern-annotell:
Documentation for feature flags and ability to opt out of motion comp.


docs-src/docs/input-api/inputs/lidars_with_imu_data.md
```diff
@@ -31,10 +28,35 @@ unit of time. Therefore all unix timestamps needs to be provided in nanoseconds.
...
 
 :::tip Use dryrun to validate input
 Setting `dryrun` parameter to true in the method call, will validate the input using the Input API but not create any inputs.
 :::
+
+## Enable/disable motion compensation
+
+By default motion compensation is performed for inputs with LIDAR pointclouds when IMU data is provided.
+
+Whether motion compensation is enabled or not is controlled by an [input feature flag](../feature_flags) which is **set** (i.e. enabled) by default. To disable motion compensation you must provide a different set of flags from the default, and not include the motion compensation flag. Calling the `create` method for an input, and not specifying feature flags *at all* is equivalent to using the defaults, and motion compensation will be performed.
```
Ragge-dev, line 59: > which is **set** (i.e. enabled) by default.

I think it is better to simply call features enabled/disabled. Otherwise we should use **unset** in some places which sounds a bit off to me

docs-src/docs/input-api/inputs/lidars_with_imu_data.md
```diff
@@ -1,9 +1,6 @@
 ---
 title: Motion Compensation for Multi-Lidar Setups
 ---
-:::info This feature is not yet available
-This feature is not currently available for public use, but will be shortly.
-:::
```
Ragge-dev, line 6: :1st_place_medal:

docs-src/docs/input-api/feature_flags.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Supported Features
+
+Individual flags are found on the enums found within `FeatureFlags`.
+
+### `PointCloudFeatures`
+
+| Flag                  | Default state | Description                                                                               |
+|:----------------------|:--------------|:------------------------------------------------------------------------------------------|
+| `MOTION_COMPENSATION` | Set           | Enable [motion compensation of point clouds](inputs/lidars_with_imu_data) using IMU data. |
```
Ragge-dev, line 28: I prefer using Enabled/Disabled instead of Set/Unset, thoughts?

Ragge-dev:
Would it also be nice to show how a user can see what feature-flags they have set, in case you want to double check, or should this be considered trivial?

docs-src/docs/input-api/feature_flags.md
petern-annotell: No strong preference either way. Could argue semantics, but not meaningfully!

PR #162, opened 2022-06-23
Int 1230 update python client

petern-annotell:
Client side support for feature flags, and in turn, for disabling motion compensation.

Depends on backend PR.


annotell-input-api/annotell/input_api/model/input/feature_flags.py
```diff
@@ -0,0 +1,39 @@
...
+    Enum type for feature flags.
+    """
+    pass
+
+
+class FeatureFlags:
+    """
+    Holds a set of flags to control optional features of the input creation process.
+    """
+    flags: Set[Flag]
```
Ragge-dev, line 18: Is this used? I thought this was only syntax used for data-classes or pydantic

annotell-input-api/annotell/input_api/model/input/feature_flags.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def __init__(self, *flags):
+        self.flags = set(flags)
+
+    @staticmethod
+    def defaults() -> FeatureFlags:
+        return FeatureFlags(FeatureFlags.PointCloudFeatures.MOTION_COMPENSATION)
+
+    @staticmethod
+    def _all_() -> set(Flag):
```
Ragge-dev, line 31: For type hinting we should use Set[Flag]

Ragge-dev: Also, why trailing _ ? From what I know prefix with _ is to indicate private method, suffix with _ is to break naming conflicts with keywords in python. So I wonder if the trailing _ is really needed

annotell-input-api/README.md
```diff
@@ -12,6 +12,12 @@ Documentation about how to use the library can be found [here](https://annotell.
 
 All notable changes to this project will be documented in this file.
 
+## [1.2.3] - 2022-06-23
```
Ragge-dev, line 15: Instead of creating a new version for every new feature, we have in the past allowed for features to be grouped in the same new realase. If we change the data to TBD we can change that data to the actual date of the realease.

However, is this something we should release asap for motional? Then you can ignore what I just said:)

Ragge-dev (changes requested):
I think you have done some great work, only thing I want to discuss is the feature_flags.py file. I would like us to be certain how we want this to be used, so that the way is paved for feature flags being added in a consistent way in the future

annotell-input-api/annotell/input_api/model/input/feature_flags.py
```diff
@@ -0,0 +1,39 @@
...
+
+from enum import EnumMeta, Flag
+from typing import Set, Dict
+
+
+class FeatureFlag(Flag):
+    """
+    Enum type for feature flags.
+    """
+    pass
```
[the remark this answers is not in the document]
Ragge-dev: I realize now that users really only interface towards the FeatureFlags method, so in this case I guess it is more how we want to work with Flags in the future:)

annotell-input-api/annotell/input_api/model/input/feature_flags.py
petern-annotell: Do you mean the comment? It's a standard docstring, other things understand them too (e.g. pycharm).

annotell-input-api/annotell/input_api/model/input/feature_flags.py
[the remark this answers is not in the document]
petern-annotell: In this case since we don't (in the end) use bitwise combinations, it can be `Enum` instead of `Flag`, yep! Holdover from before I had stubbed this all out and wasn't sure if `Feature.FOO | Feature.BAR` would be somehow useful!

The string value is an internal impl detail between the client and backend, so I think exposing that to users at all wouldn't be good.

Happy to discuss

annotell-input-api/annotell/input_api/model/input/feature_flags.py
```diff
@@ -0,0 +1,32 @@
...
+
+from enum import EnumMeta, Enum
+from typing import Set, Dict
+
+
+class FeatureFlags:
+    """
+    Holds a set of flags to control optional features of the input creation process.
+    """
+    flags: Set[Enum]
```
Ragge-dev, line 11:
```suggestion
```

Ragge-dev (approved):
:ship: :it:

PR #163, opened 2022-06-30
bugfix for openlabel section

Ragge-dev:
If you look in our public docs there are malformated in the openlabel section, quick fix to resolve that.


olinoskar (approved):
Great catch!

PR #164, opened 2022-07-04
Add repr for feature flags

petern-annotell:
Adds a repr so end users can print their flags or whatever and see what is contained in the set.


Ragge-dev (approved):
:ship: :it:

PR #165, opened 2022-07-06
Nojira fix featureflag defaults

petern-annotell:
Fixes runtime err in 1.2.3 when no flags are specified on creating an input.


Ragge-dev (approved):
Please also make sure that the tests work properly, feel like this should have been caught

PR #167, opened 2022-07-12
fixed pre-annotations test

Ragge-dev:
Some pre-annotation tests failed because we have started to validate the pre-annotation. I didn't see a good reason to keep the logic of creating an empty pre-annotation so instead we now test that we can create an input from a scene without a pre-annotation too.


annotell-input-api/examples/lidars_and_cameras_seq_with_pre_annotations.py
```diff
@@ -84,20 +85,11 @@ def run(
     # Simulate waiting for scene to be created since this is a async process
     # The rest needs to be run once the scene creation has finished
     print(f"Sleeping 10 seconds in order to wait for the scene to be created")
```
petern-annotell, line 87: Durations are out of sync 

```suggestion
    print(f"Sleeping 20 seconds in order to wait for the scene to be created")
```

petern-annotell (approved):

PR #169, opened 2022-08-10
Changed from internalId to sceneUuid on initialized input

olinoskar:


annotell-input-api/annotell/input_api/resources/abstract/creatable_input_api_resource.py
```diff
@@ -68,11 +68,11 @@ def _post_input_request(
 
             return response
 
-    def _get_input_uuid(self, body: Dict) -> Tuple[Dict, InitializedInput]:
+    def _get_scene_uuid(self, body: Dict) -> Tuple[Dict, InitializedInput]:
         json_resp = self._client.post(INPUTS_ROUTE + "/initialize")
 
         init_input = InitializedInput.from_json(json_resp)
-        body["internalId"] = init_input.internal_id
+        body["internalId"] = init_input.scene_uuid
```
petern-annotell, line 75: Just to check this `internalId` wasn't also meant to be updated?

petern-annotell (approved):
LGTM if the following question is answered!

annotell-input-api/annotell/input_api/resources/abstract/creatable_input_api_resource.py
olinoskar: I did at first but then I saw that it is used in the create-input requests which would mean that the api would also have to change

PR #170, opened 2022-08-12
Add param for compat with modern models

petern-annotell:
Trivial tricksy API compat fix for `to_dict` between legacy and modern pydantic camera calibration models. Doesn't change the behavior at all if you don't count "no longer throws an error at runtime".


olinoskar (approved):
Even though it's a bit ugly to have a parameter that is not being used, I think that this is the simplest approach. It means fewer changes when it's time to move away from the legacy approach completely

PR #172, opened 2022-08-16
Fixes

alydh-annotell:


olinoskar (approved):
Great stuff. Let's make sure that the next input api client uses (and works with) this new version

PR #180, opened 2022-09-08
Add camera calibration model

simonbastas:
Added new camera calibration model
Added test for new model


annotell-input-api/examples/calibration/create_cygnus_calibration.py
```diff
@@ -0,0 +1,23 @@
...
+    return CygnusCalibration(
+        position=camera_position,
+        rotation_quaternion=camera_rotation,
+        camera_matrix=camera_camera_matrix,
+        distortion_coefficients=camera_distortion_coefficients,
+        principal_point=principal_point,
+        distortion_center=distortion_center,
+        image_height=1080,
+        image_width=1920,
+        field_of_view=190.0
```
olinoskar, line 22: Are you sure that it should be 190.0 here?

PR #188, opened 2022-09-30
Added kognic-io

olinoskar:


petern-annotell (approved):
Can't see a reason to hold this back! Have not run tests myself but happy to ship it if run against staging.

PR #190, opened 2022-09-30
Removed deprecated get_annotations

olinoskar:
Smarteye is still using this so maybe we shouldn't do this? On the other hand, they will still be able to use the old package for some time


alydh-annotell (approved):
:heart_eyes:

PR #194, opened 2022-10-12
Added scene as a key concept and removed everything that's deprecated

olinoskar:


docs-src/docs/kognic-io/working_with_inputs.md
```diff
@@ -29,11 +29,10 @@ organization # root for projects
        └── ...
 ```
 
-There are 3 ways in order to create inputs
+There are 2 ways in order to create inputs
```
kallus, line 32:
```suggestion
There are 2 ways to create inputs
```

kallus (changes requested):
Language is easily readable, I think.

kallus (approved):

PR #195, opened 2022-10-13
Convert dataclasses to pydantic

petern-annotell:
All remaining `@dataclass`-es have been ported over to pydantic models.

Quite a lot of changes that fall into some patterns:
- Replacing the dataclass annotations by extending from `BaseSerializer`
- Removing custom `to_dict` and `from_json` impls and adding field aliases instead
- Set default values on fields that previously could be empty, so that they don't have to be specified everywhere
- Juggled a lot of imports with @olinoskar 's help; this prevented a circular dependency but means some classes that could be imported from any input model sub-package no longer can.

Clients will need to update their code. I will create a separate PR for updating the docs with a migration guide.


kognic-io/kognic/io/model/projects/project_batch.py
```diff
@@ -15,22 +13,12 @@ class ProjectBatchStatus(str, Enum):
...
-    def from_json(js: dict):
-        return ProjectBatch(
-            project=js["project"],
-            batch=js["batch"],
-            title=js["title"],
-            status=ProjectBatchStatus(js["status"]),
-            created=ts_to_dt(js["created"]),
-            updated=ts_to_dt(js["updated"])
-        )
+    # TODO for created/updated does ts_to_dt need to be given as a validator (or something) with pydantic,
```
kallus, line 24:
```suggestion
```
I think we should avoid checking in TODO lines. I suggest adding this to jira instead if it in not convenient to solve it now.

kallus (changes requested):
Very nice to see a negative change of LOC count! :)

kognic-io/kognic/io/model/projects/project_batch.py
petern-annotell: Yep good catch :+1: will nuke it

kallus (approved):

kallus (approved):
LGTM

PR #196, opened 2022-10-14
kognic-io migration guide

petern-annotell:
Adds a page to documentation about migrating from `annotell-input-api` (1.3.1) to the as-yet unreleased `kognic-io` (1.0.0) 

I also renamed the docs dir which in hindsight was not necessary; I expect it will mostly merge OK with the other PR.

This PR will need to be carefully merged with #192 and released at about the same time as #188 and the initial library version for things to make sense.


docs-src/docs/a2k_migration_guide.md
```diff
@@ -0,0 +1,69 @@
+---
+title: annotell-input-api to kognic-io
+---
+
+## Overview
+
+In September 2022 Annotell rebranded as Kognic.
+
+We took opportunity to make some changes to our public Python libraries during the transition. Some of these changes are incompatible with code written against the last version of `annotell-innput-api` (1.3.1 at time of writing).
```
olinoskar, line 9:
```suggestion
We took opportunity to make some changes to our public Python libraries during the transition. Some of these changes are incompatible with code written against the last version of `annotell-input-api` (1.3.1 at time of writing).
```

docs-src/docs/a2k_migration_guide.md
```diff
@@ -0,0 +1,69 @@
...
+
+`annotell-input-api` will now only be updated to fix critical bugs. All users are encouraged to migrate to `kognic-io` as soon as possible.
+
+## Breaking changes
+
+### New packages
+
+Packages under `annotell.import_api` were moved to `kognic.io`.
```
olinoskar, line 17:
```suggestion
Packages under `annotell.input_api` were moved to `kognic.io`.
```

olinoskar, line 17: Maybe we can add something like "and in most places it will be enough to change imports accordingly"

olinoskar (approved):
Great guide. I think that this will be very helpful for users and for us since we won't have to handle a million support issues about "I can't import this anymore". I had a few minor comments but feel free to merge when those are fixed

PR #197, opened 2022-10-14
Add principal point distortion to docs

petern-annotell:
Adds the new calibration type to the API docs. Should be merged before into #193 else change the merge base to `gh-pages`.

I expect the built files to conflict; that commit can be skipped.


kallus (approved):
Do we want to keep the text representations of tables looking nice, and do we want to keep the text width below some limit?

Other than that it looks good!

olinoskar (commented):
Why all the changes to js-files?

petern-annotell:
> Why all the changes to js-files?

That's from building the docs. I expect to discard that dir and rebuild it when all the kognic-io 1.0 stuff is merged together. I suppose I didn't need to commit it but it's nice to have a PR that works stand alone IMO.

PR #199, opened 2022-10-17
Bumped minimum python version from 3.6 to 3.7

olinoskar:


petern-annotell (approved):
Sounds like progress to me!

(for other reviewers: we released a change a while ago using something from the enum package which accidentally made us require 3.7; this just makes it official)

PR #206, opened 2022-10-19
Allow metadata to be omitted

petern-annotell:


olinoskar (commented):
maybe it's better to default it to an empty dict? Not sure how Scala will handle the None case

olinoskar (approved):

PR #209, opened 2022-10-31
Add support for passing bytes

petern-annotell:
Quite a large PR which boils down to: Add two new ways to pass binary data for images and pointclouds. This is back-compatible because it does not rely on any backend changes and doesn't remove anything.

Previously you could do this:

```py
image = Image(filename="foo.jpg", sensor_name="cam1")
```

... and we would upload the locally-resolved file `./foo.jpg`. Now you can also just give us a blob with the bytes you already have:

```py
data: bytes = b'pretend this is valid jpg data'
image = Image(filename="foo.jpg", sensor_name="cam1", data=FileData(data=data, format=FileData.Format.JPG))
```

Or give us a callback which can get the blob later, given the filename:

```py
# incrementally, from a callback:
def my_callback(name: str) -> bytes:
   ...

image = Image(filename="foo.jpg", sensor_name="cam1", data=FileData(callback=my_callback, format=FileData.Format.JPG))
```

- Filename is always required, though it's not really used with `FileData(data=...)`
- Callback arg is the file name
- Filenames we are given are always suffixed for their filetype if it's not there (`foo.jpg` is untouched, `foo.bar` becomes `foo.bar.jpg`), but only for internal use (we don't change the user's filename in the model)
- Callbacks are evaluated iteratively during the upload, i.e. we upload one file, call the next callback, etc.


kognic-base-clients/kognic/base_clients/cloud_storage/upload_handler.py
```diff
@@ -52,25 +51,45 @@ def _handle_upload_error(self, resp: Response, number_of_retries: int):
...
+
+    def _upload(self, upload_spec: UploadSpec, data: IOBase):
+        headers = {"Content-Type": upload_spec.content_type}
+        self._upload_file(upload_spec.destination, data, headers, self.max_num_retries)
+
+    def _upload_from_blob(self, upload_spec: UploadSpec):
+        log.debug(f"Blob upload for filename={upload_spec.filename}")
+        self._upload(upload_spec, upload_spec.data)
+
+    def _upload_from_local_file(self, upload_spec: UploadSpec, folder: Optional[Path]):
```
olinoskar, line 81: I wonder if the `folder` parameter can be removed here and in the methods using this one. I don't see it being used anywhere now

kognic-base-clients/kognic/base_clients/cloud_storage/upload_handler.py
```diff
@@ -52,25 +51,45 @@ def _handle_upload_error(self, resp: Response, number_of_retries: int):
...
         headers = {"Content-Type": "application/json"}
         self._upload_file(url, file, headers, self.max_num_retries)
+
+    def _upload(self, upload_spec: UploadSpec, data: IOBase):
+        headers = {"Content-Type": upload_spec.content_type}
+        self._upload_file(upload_spec.destination, data, headers, self.max_num_retries)
+
+    def _upload_from_blob(self, upload_spec: UploadSpec):
+        log.debug(f"Blob upload for filename={upload_spec.filename}")
+        self._upload(upload_spec, upload_spec.data)
```
olinoskar, line 79: The types won't be correct here. You will get `bytes` from `upload_spec.data` but the `_upload` method requires an `IOBase`. Perhaps you could convert it to a `BytesIO`?

kognic-base-clients/kognic/base_clients/cloud_storage/upload_handler.py
```diff
@@ -52,25 +51,45 @@ def _handle_upload_error(self, resp: Response, number_of_retries: int):
...
+        filename = upload_spec.filename
+        file_path = folder.joinpath(filename).expanduser() if folder else Path(filename).expanduser()
+        file = file_path.open('rb')
+        self._upload(upload_spec, file)
+
+    def _upload_from_callback(self, upload_spec: UploadSpec):
+        log.debug(f"Upload from callback for filename={upload_spec.filename}")
+        try:
+            data = upload_spec.callback(upload_spec.filename)
+            self._upload(upload_spec, data)
```
olinoskar, line 92: Same comment as above regarding the types

kognic-io/kognic/io/resources/input/file_data.py
```diff
@@ -0,0 +1,62 @@
...
+        self.format = format
+        self.data = data
+        self.callback = callback
+        self.check_data_source()
+
+    def check_data_source(self):
+        if self.data is None and self.callback is None:
+            raise ValueError(f"One of data or callback must be set on FileData")
+        if self.data is not None and self.callback is not None:
+            raise ValueError(f"Only one of data or callback may be set on FileData")
```
olinoskar, line 46: I think it would be good to check the type of `data` here

kognic-io/kognic/io/resources/input/file_data.py
```diff
@@ -0,0 +1,62 @@
...
+        @property
+        def is_pointcloud(self):
+            return self in (FileData.Format.CSV, FileData.Format.LAS, FileData.Format.LAZ, FileData.Format.PCD)
+
+        @property
+        def is_image(self):
+            return self in (FileData.Format.JPG, FileData.Format.PNG)
+
+    format: Format
+    data: Optional[bytes]
```
olinoskar, line 33: Can we use a better name than `data`, which can mean anything. I don't have any great suggestions though. Maybe `content`, but that's kind of the same. No need to change this if you can't come up with anything better

kognic-io/kognic/io/resources/input/file_data.py
```diff
@@ -0,0 +1,62 @@
+import mimetypes
+from enum import Enum
+from typing import Optional, Callable
+
+FileDataCallback = Callable[[str], bytes]
+
+
+class FileData:
+    """
+    Describes a source of file data to upload to the input API at a later point.
```
olinoskar, line 10: I have tried to avoid using `input API` in the documentation. I think that you can simply remove that here.

In addition, we should probably add a docstring here with a description about the different parameters. This is especially important on this model since users will use this directly.

kognic-io/kognic/io/resources/input/lidars.py
```diff
@@ -27,22 +27,23 @@ def create(
...
+         to all available annotation types). Passing an empty list will result in the same behaviour as passing `None`.
         :param dryrun: If True the files/metadata will be validated but no input job will be created.
         :param feature_flags Optional set of feature flags to control the input creation process.
         :returns CreateInputResponse: Class containing id of the created input job, or `None` if dryrun.
         """
-
-        imu_data = lidars.imu_data
-        payload = lidars.to_dict()
+        if not isinstance(lidars, model.Lidars):
+            raise ValueError(f"Cannot create a {type(lidars)} via this endpoint")
```
olinoskar, line 37: Please remove this. This will get the a "not supported" issue from the input api if they try to use this. We have also removed all references to this in the documentation. The reason for not removing it, is if we want to revive it at some point. Then they don't have to use another version of the client. If we want to remove support in the client directly, I think that we should remove all this instead

kognic-io/kognic/io/resources/input/lidars_sequence.py
```diff
@@ -27,22 +27,22 @@ def create(
...
+         to all available annotation types). Passing an empty list will result in the same behaviour as passing `None`.
         :param dryrun: If True the files/metadata will be validated but no input job will be created.
         :param feature_flags Optional set of feature flags to control the input creation process.
         :returns CreateInputResponse: Class containing id of the created input job, or `None` if dryrun.
         """
-
-        imu_data = lidars_sequence.imu_data
-        payload = lidars_sequence.to_dict()
+        if not isinstance(lidars_sequence, ls_model.LidarsSequence):
+            raise ValueError(f"Cannot create a {type(lidars_sequence)} via this endpoint")
```
olinoskar, line 37: Remove. See reason above

kognic-io/tests/create_input_from_bytes_test.py
```diff
@@ -0,0 +1,150 @@
...
+from tests.utils import TestProjects
+
+lidar_sensor = "lidar"
+cam_sensor = "RFC01"
+img_name = "./examples/resources/img_RFC01.jpg"
+pc_name = "./examples/resources/point_cloud_RFL01.las"
+anno_types = ['object-detection']
+
+
+class TestCreateInputFromBytes:
```
olinoskar, line 27: Perhaps you can rename this to `TestCreateSceneFromMemory` or similar since it covers both bytes and callback

olinoskar: Or maybe there should be two separate tests. One for bytes and one for callbacks?

olinoskar (changes requested):
Great job! I really like what you did with `FileData` and `UploadSpec`. I have a few comments that I think that you can have a look at. No major things though. 

Apart from the comments I also thought that about the file extensions on FileData. Currently if you pass 

```python
Image(filename="foo.bar", sensor_name="cam1", data=FileData(data=data, format=FileData.Format.JPG))
```

you get `foo.bar.jpg`. Wouldn't it be better to validate that these extensions are the same instead? Otherwise you would get `foo.png.jpg` if you pass

```python
Image(filename="foo.png", sensor_name="cam1", data=FileData(data=data, format=FileData.Format.JPG))
```

which doesn't really make sense. On the other hand, we can say that this specific use case is not allowed. Not sure what I think here...

olinoskar:
Another thing I though of now is that we could make the callback accept any key-word arguments

```python
class UploadSpec:
    data: Optional[bytes]
    callback: CallbackFunction
    callback_kwargs: dict # passed to callback
    ...

def _upload_from_callback(self, upload_spec: UploadSpec):
    # also with the try-block
    data = upload_spec.callback(upload_spec, filename, **upload_spec.callback_kwargs)
    self._upload(upload_spec, data)
```

Do what you want with this though. It could also be a nice future improvement

kognic-base-clients/kognic/base_clients/cloud_storage/upload_handler.py
petern-annotell: Will do as a separate PR just to keep this clean(ish)

kognic-base-clients/kognic/base_clients/cloud_storage/upload_handler.py
petern-annotell: Good catch, probably knew what I intended here and just didn't write it correctly! Will revisit.

kognic-io/kognic/io/resources/input/file_data.py
petern-annotell: I'm happy with data because I can't really come up with anything, and it _is_ the filedata's data

kognic-io/kognic/io/resources/input/lidars.py
petern-annotell: This is just a check that you're trying to create a lidars via the lidars endpoint, not any other kind of input -- I put it here so all the `.create` methods are consistent (then as you can say, we can revive it a bit more straight-forwardly). Is that not great?

petern-annotell:
> Another thing I though of now is that we could make the callback accept any key-word arguments

Hmmmm! I do like the idea of allowing arbitrary extra args to be passed to the callback. But, in that case, I _think_ it's possible to bake the args into the callback to achieve this... kind of how in scala if you have to do `foo(1)(2)` you could pass `foo(1)` as a callback? So the caller can create a closure over their kwargs before they give us the "rest" of the function.

```python
def my_callback(name1, named2, **kwargs):
    def inner(filename):
        # do something with filename, arg1, arg2 and kwargs.
    return inner

FileData(..., callback=my_callback)
```

kognic-io/kognic/io/resources/input/lidars.py
olinoskar: You're right. Sorry for not reading carefully

olinoskar:
> > Another thing I though of now is that we could make the callback accept any key-word arguments
> 
> Hmmmm! I do like the idea of allowing arbitrary extra args to be passed to the callback. But, in that case, I _think_ it's possible to bake the args into the callback to achieve this... kind of how in scala if you have to do `foo(1)(2)` you could pass `foo(1)` as a callback? So the caller can create a closure over their kwargs before they give us the "rest" of the function.
> 
> ```python
> def my_callback(name1, named2, **kwargs):
>     def inner(filename):
>         # do something with filename, arg1, arg2 and kwargs.
>     return inner
> 
> FileData(..., callback=my_callback("some", "args", ... )
> ```

Good idea. We could add an example in the docs about this. I am sure that someone will ask about it.

petern-annotell:
Done! See https://github.com/annotell/public-docs/pull/1

kognic-io/kognic/io/resources/input/file_data.py
petern-annotell: Turns out to be a pain since the type is now a union and cannot be used in `isinstance` in this context apparently!

kognic-base-clients/kognic/base_clients/cloud_storage/upload_handler.py
petern-annotell: #211

olinoskar (approved):
Nice!

PR #212, opened 2022-11-18
Fixed parsing of metadata types

olinoskar:


kognic-io/kognic/io/model/input/metadata/metadata.py
```diff
@@ -22,11 +23,20 @@ class MetaData(BaseModel):
     class Config:
         extra = "allow"
 
+    @root_validator
+    def validate_flat_typs(cls, values):
```
petern-annotell, line 27: Typo? 
```suggestion
    def validate_flat_types(cls, values):
```

kognic-io/kognic/io/model/input/abstract/sequence_frame.py
```diff
@@ -1,8 +1,10 @@
 from abc import ABC
 
 from kognic.io.model.input.abstract.base_frame import BaseFrame
+from kognic.io.model.input.metadata.metadata import AllowedMetaData
 
 
 class SequenceFrame(BaseFrame, ABC):
     frame_id: str
     relative_timestamp: int
+    metadata: AllowedMetaData = dict()
```
petern-annotell, line 10: Is `dict()` correct here, given `AllowedMetadata` is an alias for `MetaData` (a class)?

petern-annotell (changes requested):
Really nice that we factored out the repeated metadata fields!

Some trivia to fix.

Just to be clear, this doesn't solve the lat/long rounding issue? Or does it do it in some subtle way?

kognic-io/kognic/io/model/input/abstract/sequence_frame.py
olinoskar: It would probably be more correct to have an empty `AllowedMetData` as the default here. However, I want to make it clear that as simple dictionary can be used as well (pydantic actually parses it to a `MetaData`). Maybe it would make sense to remove `AllowedMetadata` though and simply use `MetaData`?

olinoskar:
Yes, this does address the rounding of lat/long. The reason is that pydantic tries to convert types by default. When you have union types it will begin by trying to parse it to the first one. If that fails, it will move on to the second one. If that fails, it will move on to the third one and so on. We had the following union type
```python
Union[int, float, str, bool]
```
which means that it will try `int` first, `float` second and so on. When we then have a float, as in the case with lat/long, it will succeed in parsing the value to an integer (`int(1.45675) = 1`) and hence the error. this means that we could never get a `float` (and not a `bool` either for that matter). 

What was done in this PR was to removing the parsing of types. Instead we simply validate that the value is of the correct type (`int`, `float`, `str` or `bool`), so that we don't get deeply nested dicts or lists for example

kognic-io/kognic/io/model/input/abstract/sequence_frame.py
petern-annotell: I guess my question then is what are the semantics of "allowed" when `AllowedMetaData` is used vs `MetaData`? Not worth holding the PR back tho over this IMO : )

petern-annotell (approved):

kognic-io/kognic/io/model/input/abstract/sequence_frame.py
olinoskar: I will change this to `MetaData` I think that makes the most sense

petern-annotell (approved):

PR #213, opened 2022-11-18
Created test for creating input with dropouts

olinoskar:
I was working on [this](https://annotell.atlassian.net/jira/software/projects/SP/boards/60?selectedIssue=SP-240) and figured I could push my work here


kognic-io/examples/lidars_and_cameras_seq_with_dropouts.py
```diff
@@ -0,0 +1,83 @@
...
+                    ResourceModel.PointCloud(filename=examples_path + "/resources/point_cloud_RFL01.csv", sensor_name=lidar_sensor1),
+                    ResourceModel.PointCloud(filename=examples_path + "/resources/point_cloud_RFL02.csv", sensor_name=lidar_sensor2),
+                ],
+                images=[ResourceModel.Image(filename=examples_path + "/resources/img_RFC01.jpg", sensor_name=cam_sensor1)],
+            ),
+            LCSM.Frame(
+                frame_id="2",
+                relative_timestamp=100,
+                point_clouds=[
+                    ResourceModel.PointCloud(filename=examples_path + "/resources/point_cloud_RFL02.csv", sensor_name=lidar_sensor1),
```
petern-annotell, line 51: Is it worth a comment here to say "RFL02 has dropout in this frame"?

petern-annotell (approved):

kognic-io/examples/lidars_and_cameras_seq_with_dropouts.py
olinoskar: Perhaps yes. Will make it easier for the reader

olinoskar: Or maybe not necessary. It's apparent from the name that it contains a dropout and it only contains one frame so I don't that a further explanation is needed here

PR #215, opened 2022-11-28
Retry uploads on network timeout

petern-annotell:
If an upload request failed in the HTTP phase (request timeout, 4xx, 5xx) we would handle the error and maybe retry based on what it was.

If an upload request failed prior to that, i.e. during TCP or SSL handshake, we would throw it on and not retry.

This PR adds retrying for `Timeout` errors, which shares the retry counter with HTTP errors.

This should improve our resilience against errors like these:

```
requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='[storage.googleapis.com](http://storage.googleapis.com/)', port=443): Read timed out. (read timeout=60)
```


olinoskar (approved):
Really nice! This makes total sense.
Can you also update the dependency and README in kognic-io?

PR #220, opened 2022-12-21
Support for Aggrigated Lidars And Cameras Sequence

olinoskar:
TODO:
- [ ] the tests are more or less copied from LCS, so check that they all make sense
- [ ] make sure that the tests pass


kallus (approved):
Looks good!

PR #221, opened 2023-01-16
New function get_partial_annotations

olinoskar:
I talked to Anders earlier and he expressed that he would like to be able to query finished annotations without getting the annotation contents. Since it was a rather small change I created this PR. We can then discuss whether to merge it or not.


petern-annotell (approved):
Seems like a nice idea! Do we need to advertise it via docs?

PR #223, opened 2023-01-30
Validate no 3D geometry with source-specific properties in pre-annotations

petern-annotell:
May not use source-specific properties on 3d geometry for now (no known valid UC)


kallus (approved):
LGTM

PR #227, opened 2023-02-23
More robust branch checks for upload data source

petern-annotell:
Previously if the data or callback could be coerced to False we could go down the wrong branch. Since these are Optionals they'll be None if not set.

Draft PR as I've not yet tested this.


luca-caltagirone (approved):
LGTM!

PR #228, opened 2023-03-08
Sp 442 add new model fused cylindrical in kognic io

luca-caltagirone:
Added support for Fused Cylindrical camera model


olinoskar (approved):
LGTM!
Don't forget to update the changelog in the README!

kognic-io/kognic/io/model/calibration/camera/fused_cylindrical_calibration.py
```diff
@@ -0,0 +1,16 @@
...
+
+class CutAngles(BaseSerializer):
+    upper: float
+    lower: float
+
+
+class FusedCylindricalCalibration(BaseCameraCalibration):
+    calibration_type = CalibrationType.FUSEDCYLINDRICAL.value
+    cut_angles_degree: CutAngles
+    vertical_fov_degree: Optional[float] = 90
```
olinoskar, line 15: Can these be `None` or are they optional in the sense that they have a default value?

olinoskar: Also, can we rename it to vertical_fov_degree**s** instead? And the same with horizontal_fov_degree

luca-caltagirone: Yes, exactly. They have default values but the client could override them in the future if needed

PR #229, opened 2023-03-14
Update camera image sample code to use other image formats

petern-annotell:
Previously only used JPGs, now it uses JPG, PNG and the 3 kinds of WebP for different frames.

Also fixes a persistently failing test which did not wait for status to change when invalidating inputs.


kognic-io/tests/invalidate_inputs_test.py
```diff
@@ -44,6 +42,12 @@ def test_invalidate_inputs(self, client: IOC.KognicIOClient):
 
         invalidate_inputs_example.run(client=client, input_uuid=input_uuid, invalidated_reason=InvalidatedReasonInput.BAD_CONTENT)
 
-        invalidated_input = get_inputs_example.run(client=client, input_uuids=[input_uuid])[0]
+        invalidated_inputs = None
+        for _ in range(3):
```
olinoskar, line 46: Invalidation should be done synchronously so why try several times?

petern-annotell: Answer: scene invalidation is immediate but input invalidation requires a pub/sub cycle

PR #231, opened 2023-03-21
Sp 497 function to create scenes and inputs

olinoskar:
Approach 2 of interface for creating scenes/inputs with one function (Approach 1 [here](https://github.com/annotell/annotell-python/pull/230)). The difference with this one is that scenes are handled as they are completed rather than being blocked while waiting for all other scenes. This will be faster and instead of returning one container with the summary of the entire process we yield creation results for each scene/input

```python
class CreationResult(BaseModel):
    external_id: str
    scene_uuid: str
    input_uuid: Optional[str] = None  # Only defined if an input was created
    status: str
    error: Optional[CreationError] = None  # Only defined if an error occured

    def combine(self, other: CreationResult) -> CreationResult:
        raise NotImplemented
```

We could write a wrapper function where we print a summary as well, but I guess that would rather be a nice to have.

Note that we need to be very careful about how we deal with the uuids. I am not sure whether there is a way that we can make this backwards compatible in the future.


kognic-io/kognic/io/tools/scene_input.py
```diff
@@ -0,0 +1,115 @@
...
+        yield input_result
+
+
+def _start_scene_jobs(
+    resource: CreateableIOResource,
+    scenes: List[Scene],
+    dryrun: bool = False,
+    feature_flags: Optional[FeatureFlags] = None
+) -> List[SceneUuid]:
+    # Starts scene jobs. Don't catch validation errors. The returned list must have the same order as the scenes
```
petern-annotell, line 72: > Don't catch validation errors

What happens if  there are multiple?

petern-annotell: Thinking about it more, why not use CreationResults here too? Have an extra state for those that are invalid

olinoskar: An error will be raised and you will only see that one. I suspect that you will get this error for most of the scenes you're trying to create though.

olinoskar: Due to my above comment I don't think we want to continue with the other steps

kognic-io/kognic/io/tools/scene_input.py
```diff
@@ -0,0 +1,115 @@
...
+    raise NotImplemented
+
+
+def _wait_for_scene_jobs(
+    client: KognicIOClient,
+    scene_uuids: List[SceneUuid],
+    wait_timeout: int,
+) -> Generator[CreationResult, None, None]:
+    # Wait for scene jobs to complete. Yield creation results as they are or the process times out
+    # Having some good wait strategy here will be important. Think about big vs small scenes and the number of scenes
```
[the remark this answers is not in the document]
olinoskar: In a perfect world scene jobs would not get stuck in processing which means that we wouldn't need a timeout at all

kognic-io/kognic/io/tools/model.py
```diff
@@ -0,0 +1,53 @@
...
+    pre_annotation: Optional[OpenLabelAnnotation] = None
+
+
+class CreationStage(enum.Enum):  # Some suitable enum describing the stage/step where the failure happened
+    SCENE = "scene"
+    PRE_ANNOTATION = "pre-annotation"
+    INPUT = "input"
+
+
+class CreationStatus(enum.Enum):  # Some suitable enum describing the stage/step where the failure happened
```
AndersNordmark, line 28: Remove comment

kognic-io/kognic/io/tools/model.py
```diff
@@ -0,0 +1,53 @@
...
+
+Scene = TypeVar("Scene", bound=BaseInput)
+
+
+class SceneInput(BaseModel):  # Better name?
+    scene: Scene
+    pre_annotation: Optional[OpenLabelAnnotation] = None
+
+
+class CreationStage(enum.Enum):  # Some suitable enum describing the stage/step where the failure happened
```
AndersNordmark, line 22: Remove comment

kognic-io/kognic/io/tools/model.py
```diff
@@ -0,0 +1,53 @@
...
+from pydantic import BaseModel
+
+
+SceneUuid = str
+InputUuid = str
+
+Scene = TypeVar("Scene", bound=BaseInput)
+
+
+class SceneInput(BaseModel):  # Better name?
```
AndersNordmark, line 17: Suggestions:
- InputWithPreannotation
- SceneWithPreannotation
- Input
- SceneData
- AnnotationInput
-

kognic-io/kognic/io/tools/scene_input.py
```diff
@@ -0,0 +1,147 @@
...
+        if input_result.status == CreationStatus.CREATED:
+            pre_annotation = scenes_map[input_result.scene_uuid].pre_annotation
+            if pre_annotation:
+                pre_anno_result = _create_pre_annotation(client, input_result.scene_uuid, pre_annotation)
+                input_result.update(pre_anno_result)
+
+        if input_result.status == CreationStatus.CREATED:
+            input_from_scene_result = _create_input_from_scene(resource, input_result.scene_uuid, project, **kwargs)
+            input_result.update(input_from_scene_result)
+        yield input_result
```
AndersNordmark, line 63: Why are we yielding here? Isn't it more user friendly to populate a list of creation statuses in the loop and returning that?

luisf-annotell: By yielding, we enable the user to trigger other processes as the inputs are completed, rather than having to wait for all of them to finish.

kognic-io/kognic/io/tools/scene_input.py
```diff
@@ -0,0 +1,147 @@
...
+    return [resource.create(scene, dryrun=dryrun, feature_flags=feature_flags).input_uuid for scene in scenes]
+
+
+def _wait_for_scene_jobs(
+    client: KognicIOClient,
+    scene_uuids: List[SceneUuid],
+    wait_timeout: float,
+    sleep_time: float
+) -> Generator[CreationResult, None, None]:
+    def poll_input_status(su: str) -> CreationResult:
```
olinoskar, line 82: How would this work in practice? Would we wait for 5 scenes at a time or would we more or less wait for all scenes at the same time? In case of the latter, would we fire lots of calls to `get_inputs_by_uuids` at the same time?

It would maybe be nice to use the fact that `get_inputs_by_uuids` actually takes a list of uuids. Can we do that here? The things is that that endpoint is quite heavy so it would be nice of we could avoid bombarding it ;)

kognic-io/kognic/io/tools/scene_input.py
```diff
@@ -0,0 +1,147 @@
...
+    :param wait_timeout: Time to wait for scene jobs to complete.
+    :param kwargs: Passed to the `create_from_scene` method. For example batch and annotation_types
+    :returns CreationResult: Contains information about the creation process. For example, whether it succeeded or failed
+    """
+
+    scenes = [si.scene for si in scene_inputs]
+    resource = _get_creatable_resource(client, scenes)
+    scene_uuids = _start_scene_jobs(resource, scenes, dryrun=dryrun, feature_flags=feature_flags)
+
+    if dryrun:  # Don't do anything more if dry-run
```
olinoskar, line 49: This is a minor detail, but I realised that this will break things for the user if they have `dryrun=True` and they expect a generator

.gitignore
```diff
@@ -137,3 +137,6 @@ dmypy.json
 
 # IDE folders
 *.vscode/
+
+# Mac garbage
+.DS_Store
```
olinoskar, line 142: Love it!

kognic-io/kognic/io/tools/model.py
olinoskar: Out of these I like `SceneWithPreannotation` the best

kognic-io/kognic/io/tools/scene_input.py
```diff
@@ -0,0 +1,147 @@
...
+        for future in as_completed(features):
+            yield future.result()
+
+def _create_pre_annotation(client: KognicIOClient, scene_uuid: str, pre_annotation: OpenLabelAnnotation) -> CreationResult:
+    # Try to create a pre-annotation and return a creation result
+    try:
+        client.pre_annotation.create(scene_uuid, pre_annotation, dryrun=False)
+        return CreationResult(scene_uuid=scene_uuid, status=CreationStatus.CREATED)
+    except HTTPError as error:
+        logger.error(f"Failed to create pre-annotation for scene {scene_uuid}. {error}")
```
olinoskar, line 112: This will make errors clearer. It's not the full stacktrace but it's a bit more than when parsing to a string
```suggestion
        logger.error(f"Failed to create pre-annotation for scene {scene_uuid}. {error}", exc_info = error)
```

kognic-io/kognic/io/tools/scene_input.py
```diff
@@ -0,0 +1,147 @@
...
+        return CreationResult(scene_uuid=scene_uuid, status=CreationStatus.FAILED, error=creation_error)
+
+
+def _create_input_from_scene(resource: CreateableIOResource, scene_uuid: SceneUuid, project: str, **kwargs) -> CreationResult:
+    # Create input from scene. Catch errors and put them in the result
+    try:
+        resource.create_from_scene(scene_uuid, project, **kwargs)
+        return CreationResult(scene_uuid=scene_uuid, input_uuid=scene_uuid, status=CreationStatus.CREATED)
+    except HTTPError as error:
+        logger.error(f"Failed to create input from scene {scene_uuid}. {error}")
```
olinoskar, line 123:
```suggestion
        logger.error(f"Failed to create input from scene {scene_uuid}. {error}", exc_info=error)
```

olinoskar (commented):
Overall a nice and elegant PR. I would like to see if the fact that `get_inputs_for_uuids` takes multiple uuids can be leveraged to reduce the load on the api though. It would also be nice to understand how the current solution works. 

One question I have is whether this has been tested and in that case how. Preferably it would be tested with a large number of scenes with pre-annotations and also scenes that are quite big in size. By a large number I mean one scene created many times.

PR #234, opened 2023-03-31
Sp 497 function to create scenes and inputs approach 3

luisf-annotell:
Reimplements the scene job waiting process + some cleanup.


kognic-io/kognic/io/tools/scene_input.py
```diff
@@ -79,29 +79,36 @@ def _wait_for_scene_jobs(
...
-    # Wait for scene jobs to complete. Yield creation results as they are or the process times out
-    # Having some good wait strategy here will be important. Think about big vs small scenes and the number of scenes
-    with ThreadPoolExecutor(max_workers=5) as executor:
-        features = [executor.submit(poll_input_status, su) for su in scene_uuids]
-        for future in as_completed(features):
-            yield future.result()
+    remaining_scenes_uuids = set(scene_uuids)
+    start_time = time.time()
+    while time.time() - start_time < wait_timeout:
+        scenes = client.input.get_inputs_by_uuids(remaining_scenes_uuids)
```
olinoskar, line 85: I think that we need to limit the number of uuids we send here since this route tends to fail with many uuids. I think that a around a 100 should be a good number

kognic-io/kognic/io/tools/scene_input.py
```diff
@@ -79,29 +79,36 @@ def _wait_for_scene_jobs(
...
+
+        # Yield completed scenes
+        for scene_uuid in set(remaining_scenes_uuids).difference(set(s.uuid for s in scenes)):
+            yield CreationResult(scene_uuid=scene_uuid, status=CreationStatus.CREATED)
+
+        remaining_scenes_uuids = []
```
olinoskar, line 91: Why not use a set instead? Then you won't have to convert it to a set on line 88

olinoskar (changes requested):
I don't think we can fetch all scenes at once so that needs to be changed. Otherwise it looks good

kognic-io/kognic/io/tools/scene_input.py
luisf-annotell: Oops

kognic-io/kognic/io/tools/scene_input.py
```diff
@@ -79,29 +79,36 @@ def _wait_for_scene_jobs(
...
+        for scene in scenes:
+            if scene.status is InputStatus.Failed:
+                yield CreationResult(
+                    scene_uuid=scene.uuid,
+                    status=CreationStatus.FAILED,
+                    error=CreationError(stage=CreationStage.SCENE, message=scene.error_message)
+                )
+            else:
+                remaining_scenes_uuids.append(scene.uuid)
```
AndersNordmark, line 100: Won't the completed scenes also be appended here? Which means that they will be polled and returned again? Or am I missing something?

kognic-io/kognic/io/tools/scene_input.py
```diff
@@ -73,35 +73,63 @@ def _start_scene_jobs(
...
+    start_time = time.time()
+    while time.time() - start_time < wait_timeout:
+        for chunk in _chunkify_set(remaining_scenes_uuids, 100):
+            scenes = client.input.get_inputs_by_uuids(chunk)
+
+            completed = []
+            # Yield created scenes
+            for scene_uuid in chunk.difference(set(s.uuid for s in scenes)):
+                yield CreationResult(scene_uuid=scene_uuid, status=CreationStatus.CREATED)
+                completed.append(scene_uuid)
```
olinoskar, line 107: En mindre detalj kanske men du kan egentligen bara köra `remaining_scenes_uuids.remove(scene_uuid)` här så kan du droppa `completed` helt och hållet

luisf-annotell: Hmm hade nåt sånt från början tror jag, minns inte varför jag ändrade...

kognic-io/kognic/io/tools/scene_input.py
```diff
@@ -73,35 +73,63 @@ def _start_scene_jobs(
     return [resource.create(scene, dryrun=dryrun, feature_flags=feature_flags).input_uuid for scene in scenes]
 
 
+def _chunkify_set(origin_set: set, chunk_size: int) -> List[set]:
+    count = 0
```
olinoskar, line 77: This can be shortened quite a lot with a one liner. Try this
```python
return [origin_list[i:i+chunk_size] for i in range(0, len(origin_list), chunk_size)]
```
The function would then accept a list intead and return a list of lists instead

olinoskar (changes requested):
I think that the logic is much better now and we shouldn't run into issues with the input-api. There are still some minor changes that can be done though to make the code a bit prettier

kognic-io/kognic/io/tools/scene_input.py
luisf-annotell: Huh, I thought I'd get an error if I tried an index larger than the list length. But even so, I'd rather keep this approach that works with sets, saves me having to convert back and forth in the main code.

olinoskar: You don't have to convert back and forth in the main method. Just in the call to this method. And if you don't want to do that you can simply add an extra line to this function and convert from set to list

kognic-io/kognic/io/tools/scene_input.py
```diff
@@ -73,42 +74,61 @@ def _start_scene_jobs(
...
-    # Wait for scene jobs to complete. Yield creation results as they are or the process times out
-    # Having some good wait strategy here will be important. Think about big vs small scenes and the number of scenes
-    with ThreadPoolExecutor(max_workers=5) as executor:
-        features = [executor.submit(poll_input_status, su) for su in scene_uuids]
-        for future in as_completed(features):
-            yield future.result()
+    remaining_scenes_uuids = [*scene_uuids]
+    start_time = time.time()
+    chunk_size = NORMAL_CHUNK_SIZE
+    while time.time() - start_time < wait_timeout:
```
olinoskar, line 89: A very minor thing that you can address if you want to
```suggestion
    while time.time() - start_time < wait_timeout and len(remaining_scenes_uuids) > 0:
```

kognic-io/kognic/io/tools/scene_input.py
```diff
@@ -10,19 +9,21 @@
     AggregatedLidarsAndCamerasSequence, InputStatus
 from kognic.io.model.input.feature_flags import FeatureFlags
 from kognic.io.resources.abstract import CreateableIOResource
-from kognic.io.tools.model import SceneInput, CreationResult, CreationError, CreationStage, CreationStatus, Scene, SceneUuid
+from kognic.io.tools.model import SceneWithPreannotation, CreationResult, CreationError, CreationStage, CreationStatus, Scene, SceneUuid
 
 logger = logging.getLogger(__name__)
 
+NORMAL_CHUNK_SIZE = 1000
```
olinoskar, line 16: I don't think that 1000 is ever going to work actually. Maybe 100 is more reasonable
```suggestion
NORMAL_CHUNK_SIZE = 100
```

kognic-io/tests/tools_scene_input_test.py
```diff
@@ -0,0 +1,25 @@
...
+
+import kognic.io.client as IOC
+import kognic.io.model as IAM
+import kognic.openlabel.models as OLM
+from kognic.io.tools import CreationStatus
+import examples.tools_scene_input as tools_scene_input_example
+from tests.utils import TestProjects
+
+
+class TestLidarsAndCamerasSeqWithPreAnnotations:
```
olinoskar, line 13: Rename class

olinoskar (approved):
There are still a few minor things that I would like you to change but feel free to merge when that's done. Great job!

kognic-io/kognic/io/tools/scene_input.py
luisf-annotell: Oops, used it for testing, forgot to change back.

PR #236, opened 2023-04-11
Setup for version 1.1.9 of kognic-io

olinoskar:


kognic-io/setup.py
```diff
@@ -46,7 +46,8 @@
         'python-dateutil',
         "pydantic",
         "pyhumps",
-        "Deprecated"
+        "Deprecated",
+        "tqdm"
```
petern-annotell, line 50: Where was this used?

petern-annotell (changes requested):
One tiny thing but otherwise looks good! These names are much better! :)

kognic-io/setup.py
olinoskar: It was added in a previous PR

PR #239, opened 2023-04-14
Fixed bug with calibration types on calibration models

olinoskar:
The problem was that calibration models used calibration types differently, see below. This would lead to checks such as `if calibration.calibration_type == CalibrationType.TheType` to work for some models but others would require `if calibration.calibration_type == CalibrationType.TheType.value`.

To prevent this from happening when we add new models, I added `calibration_type` to the base classes. This requires the types to be the same

```python
class Calibration1(BaseModel):
   calibration_type: CalibrationType.TheType

class Calibration2(BaseModel):
   calibration_type: CalibrationType.TheType.value
```


luca-caltagirone (approved):
LGTM!

PR #243, opened 2023-05-04
Add cylindrical camera calibration

kallus:
I think it is best to add a new calibration (and projection in other repos) instead of reusing the fused cylindrical.


olinoskar (changes requested):
We have some tests for calibrations, including creating calibrations and serializing/deserializing them, and it would be great if you could add this to those tests as well (and then run them against staging). Otherwise I have no objections

kallus:
Moved this here instead https://github.com/annotell/kognic-io-python/pull/1

PR #245, opened 2023-05-10
Remove kognic-base-clients

olinoskar:
kognic-base-clients has moved to [kognic-base-clients-python](https://github.com/annotell/kognic-base-clients-python)


kallus (approved):
Nice!

PR #247, opened 2023-05-22
Remove kognic-io after moving it to a separate repo

olinoskar:


petern-annotell (approved):
-7,121 lines of LGTM :)

We can also remove the `docs-src` I think too?

olinoskar:
> -7,121 lines of LGTM :)
> 
> We can also remove the `docs-src` I think too?

We should be more careful about `docs-src` since it routes old view links in the docs to new ones (annotell -> kognic)

